PR #65, opened 2026-06-16
perf(3d): add toggleable dev HUD to measure CPU-sync vs GPU draw calls

diegokingston:
Step 1 of the 3D render-perf investigation: instrument before optimizing. A small overlay in Viewport3D (enable with ?perf, localStorage.stabileo_perf='1', or toggle live with Shift+P) reports per ~250ms window:
- fps + render ms (time spent in renderer.render)
- syncMs — CPU time in the scene-sync wrappers (syncNodes/Elements/Supports/ Loads/Shells/LocalAxes/MemberOffsets/ShellOffsets), via a perfTimed() wrapper
- draw calls + triangles (renderer.info.render, this-frame counts)
- geometry + texture counts (renderer.info.memory) — churn here during an edit exposes teardown/rebuild CPU cost

Interpretation: high `calls` + stable `geos` = GPU draw-call bound (→ shells/ sections batching); `geos`/`texs` spiking + high `syncMs` while dragging = CPU teardown/rebuild churn (→ granular versioning + drag preview). Zero cost when off (perfTimed early-returns; the render-side block is guarded by perfHud.on). Dev- only measurement tool; nothing rendered in production unless explicitly enabled.


Batuis:
**Nice work — this is a useful PR.** I ran it standalone and integrated on top of PR13 (the Basic Despiece/Joints branch) and the perf instrumentation is genuinely helpful. The HUD makes it easy to see CPU-sync vs GPU draw-call/triangle cost, and **Smooth Orbit materially reduces render load during camera motion**. Bauti confirmed it improves orbit behavior once enabled.

### Measured render-load reduction (during orbit)
These are *render-load* numbers (draw calls / triangles) from our audit — actual FPS will depend on the real GPU/CPU, but the load drop is large:

- **7-story RC building (PRO):** full-detail idle **878 draw calls / 53,946 tris** → Smooth Orbit during motion **34 draw calls / 1,282 tris**.
- **Geodesic dome:** **124,396 → 5,772 tris** during orbit.
- **La Bombonera:** **368,642 → 14,870 tris** during orbit.

### Main UX issue — it looks like a no-op until you find the toggle
The biggest friction: **Smooth Orbit is off by default and isn't persisted** (it resets to off on every reload). We initially thought PR65 "did nothing" on the 7-story model — until we realized the checkbox simply wasn't enabled, so the viewport was orbiting at full detail. Two suggestions:
- **Persist `smoothOrbit3D`** alongside the other 3D settings, so it survives reloads.
- Consider **auto-enabling / auto-activating** it for heavy / PRO / CAD models (e.g. gate on the existing `isHeavyModel` heuristic), so users get the benefit without having to discover the flag.

### Remaining CPU-side note (not a PR65 issue)
On the 7-story PRO model, even after the GPU draw load drops, some orbit lag can remain — it's main-thread bound there, not GPU bound. During a real orbit we measured **zero scene-sync rebuilds and hover suppressed**, so the residual cost is the per-frame render loop + `handleMouseMove`, which runs a ground-plane raycast and a reactive `setMouse()` status-bar update **on every pointer move with no orbit guard**. A cheap win (orthogonal to this PR, probably best on main): **skip the mouse-coordinate ground raycast / `setMouse()` while OrbitControls is actively orbiting**, keeping hover/select unchanged when not orbiting.

### Integration with PR13
PR65 and PR13 **integrate cleanly** — the only merge conflict was a small `Viewport3D.svelte` `renderOnce` hunk (PR13's Despiece pull-apart block vs your `_perfT0`), trivially resolved by keeping both. Two things to preserve when you rebase after PR13 lands:
- **PR13's Despiece LOD guard** (`setLowDetail` early-returns while `diagramType === 'despiece'`) — without it, Smooth Orbit would flash the original un-separated members over the Free-body overlay. It composes fine: PR13's guard sits ahead of your `smoothOrbit3D` check, so LOD stays disabled in Despiece.
- **PR13's `elements-batched.ts` render-cap fix** (recreate the batched geometry on grow) — needed so larger models render fully after a smaller one.

### Verdict
Strong **render-load reduction** and a good **diagnostic** PR — happy to see it land (rebased on main after PR13). It doesn't fix *all* PRO orbit lag on its own (some of that is the main-thread path above), so I wouldn't frame it as a complete fix — but persisting the setting / sensible default behavior would make the win far easier to actually experience. 👍

Batuis (approved):
Approved after local A/B testing against the merge-base. The improvement in 3D responsiveness is very considerable and worth landing.

The audit did identify follow-up items that should remain tracked, especially the visibleLoadCases3D null/empty-array signature collision that makes All/None leave stale load arrows. Other follow-ups include guarding Shift+P while typing, avoiding HUD reactivity affecting its own measurements, and clarifying Smooth Orbit availability/persistence.

These issues are acknowledged; the measured and perceived rendering improvement receives my green light.

PR #69, opened 2026-07-24
Fix/solver input validation hardening

diegokingston:


Batuis:
Diego — heads up before this goes anywhere: I did not audit the solver logic here, so it needs an engine owner.

**Why it's outside what I reviewed**

Ten of the eleven changed files are Rust solver work: eight files under `engine/src/solver/` and two solver test files. The remaining change is documentation. There is no web or product code to look at, so it falls outside the product/web review boundary I work within. More importantly, it changes *which inputs the solver accepts or rejects* — that is a behavioural change to what the tool will analyse, not just defensive hardening.

**What the reviewer will need from you**

The PR description was empty when I inspected it. For a validation change that is the critical missing piece, because there is no way to tell an intended rejection from an accidental one. Please document:

- every new rejection condition
- the exact error string each one produces
- whether any previously accepted model now fails, and if so which

**Coordination with the other engine PRs**

This touches `solver/linear.rs`, `solver/modal.rs` and `solver/staged.rs`, which the #70 → #71 → #72 stack also modifies; `solver/harmonic.rs` and `solver/modal.rs` overlap #73; `solver/linear.rs` and `solver/time_integration.rs` overlap #74. `solver/linear.rs` and `solver/modal.rs` are each rewritten by three independent PRs. It is worth agreeing a merge order across #69–#74 now rather than resolving the same conflicts three separate times in safety-critical files.

To be explicit: I verified the file scope, branch topology, and overlaps only. I reached no conclusion about numerical or solver correctness.

PR #70, opened 2026-07-24
perf(engine): assemble + factorize once for multi-case static solves

diegokingston:
solve_multi_case_2d/3d ran a full solve_* per load case (assembly + validation + factorization each) even though K is identical across cases — only the load vector changes. At 4/16 cases on a 3D frame (nf=240) the naive loop costs 3.9/16.1ms vs 1.6/2.8ms now (2.4x/5.8x, bench multi_case), and the waste grows linearly with case count.

- linear.rs: split solve_2d/solve_3d into prepare_static_* (DOF numbering, validation, stiffness-only assembly, block extraction with K_fr*u_r, conditioning, one factorization — dense Cholesky / sparse symbolic+numeric with the diagonal-shift regularization retry / LU fallback) and solve_loads(&loads) (validate case loads, rebuild only the load vector, apply prescribed-displacement RHS correction per case, triangular solve with the stored factor, identical postprocess). Public API unchanged.
- assembly.rs / sparse_assembly.rs: split each fused assembler into a K part
  + a load-vector builder (thin composition preserves bit-identical outputs); inclined transforms split into K-rotation and f-rotation helpers.
- lu.rs: lu_solve = lu_factor + lu_apply.
- load_cases.rs: prepare once, solve_loads per case; drop per-case model clones and per-combination results.clone() (new combine_results*_refs). 3D with constraints keeps the legacy per-case path.
- tests/integration/multi_case_parity.rs: exact == parity (every f64 of displacements/reactions/element forces/quad stresses) vs N full solves, dense+sparse x 2D+3D, thermal/prescribed/point loads; 1-case == single solve. Full suite green (5925 passed); clippy clean; parallel-feature lib tests green.

### DIFF
--- a/engine/benches/solver_bench.rs
+++ b/engine/benches/solver_bench.rs
@@ -1155,6 +1155,89 @@ fn bench_staged(c: &mut Criterion) {
     group.finish();
 }
 
+// ─── Multi-case load-case factorization reuse (3D frame, sparse path) ─────
+
+fn make_multi_case_input_3d(base: &SolverInput3D, n_cases: usize) -> dedaliano_engine::solver::load_cases::MultiCaseInput3D {
+    use dedaliano_engine::solver::load_cases::*;
+    let load_cases: Vec<LoadCase3D> = (0..n_cases)
+        .map(|k| {
+            let scale = 1.0 + k as f64 * 0.25;
+            let loads: Vec<SolverLoad3D> = base
+                .loads
+                .iter()
+                .map(|l| match l {
+                    SolverLoad3D::Nodal(nl) => SolverLoad3D::Nodal(SolverNodalLoad3D {
+                        node_id: nl.node_id,
+                        fx: nl.fx * scale,
+                        fy: nl.fy * scale,
+                        fz: nl.fz * scale,
+                        mx: nl.mx,
+                        my: nl.my,
+                        mz: nl.mz,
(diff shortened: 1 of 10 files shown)


Batuis:
Short version for Diego: this is the bottom of the #70 → #71 → #72 stack, so it has to be reviewed and settled before its children. I only checked scope and topology — the solver behaviour itself needs an engine owner.

**Scope**

Ten files, all under `engine/`. The substance is assembling and factorizing once for multi-case static solves, which is about as central as solver code gets. That needs two kinds of evidence before it lands: numerical regression showing results are unchanged against the current solver, and actual measurements backing the performance rationale.

**Why the order matters**

#71 and #72 are stacked on this branch. Anything corrected here propagates into both, so reviewing them before #70 is settled means reviewing them twice.

**Cross-stack overlap**

Independent of the stack, `solver/modal.rs` is shared with #73, and `postprocess/combinations.rs` plus `solver/linear.rs` with #74. #69 also touches `solver/linear.rs`, `solver/modal.rs` and `solver/staged.rs`. Once the merge order for #69–#74 is agreed, whichever unit lands later should be rebased and validated against the solver state that will actually precede it. The important part is avoiding numerical validation against a temporary base that will be replaced before merge.

Status from my side: topology and file scope verified, solver correctness not reviewed.

PR #71, opened 2026-07-24
perf(engine): reuse one factorization for moving loads and influence …

diegokingston:
…lines

Both workflows ran a full solve_* per train position / per sampled point (assembly + validation + factorization each). They now prepare once via the prepared-static API and only rebuild the RHS per position/point:

- moving_loads.rs: prepare_static_2d/3d once; per-position loads via new public moving_loads_at_position_2d/3d + solve_loads. Silent-skip and zeroed-envelope semantics preserved exactly; constrained models keep the legacy per-position loop.
- postprocess/influence.rs: one factorization + one solve_loads per sampled unit load (new influence_unit_loads_2d/3d helpers); per-point failure still yields a 0.0 ordinate. Muller-Breslau adjoint redesign noted as follow-up.

Parity: tests/integration/moving_influence_parity.rs — exact == on every f64 (envelopes, ordinates, num_positions) vs the legacy loop, 2D+3D, default and parallel builds; full suite 5925 green; clippy clean. Bench (moving_influence_reuse group): 2.1x moving 2D dense, 3.2x sparse, 3.9x influence dense, 6.4x influence sparse.

### DIFF
--- a/engine/benches/solver_bench.rs
+++ b/engine/benches/solver_bench.rs
@@ -1238,6 +1238,107 @@ fn bench_multi_case(c: &mut Criterion) {
     group.finish();
 }
 
+// ─── Moving loads / influence line factorization reuse ─────
+
+fn bench_moving_influence_reuse(c: &mut Criterion) {
+    use dedaliano_engine::postprocess::influence::{
+        compute_influence_line, influence_unit_loads_2d, InfluenceLineInput,
+    };
+    use dedaliano_engine::solver::moving_loads::{build_load_path, moving_loads_at_position_2d};
+
+    let mut group = c.benchmark_group("moving_influence_reuse");
+    group.sample_size(20);
+
+    let train = LoadTrain {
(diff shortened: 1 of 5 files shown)


Batuis:
Quick note for Diego: this is the second layer of the stack. Review it as the diff against #70, not against `main` — a cumulative view will show #70's changes as though they were introduced here.

Base is `perf/engine-multi-case-factor-reuse` (#70) at `ffbea0686641`, which is correct.

**What needs engine-owner validation**

The change reuses a single factorization across moving-load and influence-line evaluations. The engine review should define an acceptable numerical tolerance and compare it with the existing per-evaluation path, verify determinism across repeated runs on identical input, and reproduce the performance claim on representative models.

**Sequencing**

If #70 needs corrections, they have to land there first and this branch be restacked on the corrected version. Validating #71 against the current #70 would be throwaway work.

I did not review solver correctness — this is scope and sequencing only.

PR #72, opened 2026-07-24
perf(engine): sparse-native 2D linear solve and 2D modal for nf >= 64

diegokingston:
The 2D sparse path was dense everywhere except the factorization: dense n*n K assembly, O(n^2) dense->CSC conversion, dense block extraction, ~5*n^2 f64 peak memory. Route it through a new triplet assembler:

- sparse_assembly.rs: assemble_stiffness_sparse_2d — triplet K assembly mirroring dense feature-for-feature (Timoshenko shear, truss/cable, connectors, rotated springs, artificial-DOF guards, inclined transforms, connectors), producing k_ff + k_full CSC directly.
- linear.rs: prepare_static_2d uses it at nf >= 64 — symbolic+numeric Cholesky with the 3D-style regularization retry, cross-block reactions, sparse residual + refinement; dense LU fallback preserved. Dense path for nf < 64 untouched; legacy dense path kept as test oracle.
- modal.rs: 2D modal uses sparse K + sparse shift-invert op at nf >= 64.

Correctness fix included: the triplet inclined-transform (pre-existing 3D apply_inclined_transform_triplets_k and the new 2D variant) double-counted the block diagonal, corrupting k_full cross-block entries (reactions, K_fr*u_r coupling) and k_ff with nonzero block off-diagonals — a latent 3D sparse-solver bug for inclined supports. Fixed by storing full blocks and pushing each unordered pair once; dense==sparse K verified exactly.

Parity: tests/integration/sparse_2d_parity.rs — sparse vs dense oracle, rel 1e-8 statics (measured ~1e-12) / 1e-6 frequencies, default+parallel builds; suite 5925 green; clippy clean.
Bench (sparse_2d_assembly): 2.5x at nf=3600, K memory ~518MB -> ~0.5MB.


Batuis:
Diego — this is the top of the #70 → #71 → #72 stack, and its `web` CI job is currently failing. Worth sorting that out before anyone spends review time here.

**CI as of this comment**

`web` fail, `lint` pass, `test` pass. An engine-only PR failing the web job needs diagnosis before review proceeds. It may indicate a WASM or frontend-compatibility problem, but the cause was not established during this scope inspection.

**How to review it**

Incrementally against #71 (`perf/engine-moving-influence-reuse` at `f77a7de169e8`), not as a cumulative diff. The content is sparse-native 2D assembly plus a sparse path for 2D modal above a DOF threshold. Routing a solve through a different numerical path needs engine-owner review and regression evidence around the threshold. Models just below and above it should be compared with the existing/reference path using documented tolerances, including a check that switching paths does not introduce an unexplained discontinuity.

**Sequencing**

Anything corrected in #70 or #71 means restacking this rather than resolving the combined solver changes here in one pass.

No solver-correctness audit was performed on my side.

PR #73, opened 2026-07-24
perf(engine): sparse (CSC) mass matrix on the sparse eigen paths

diegokingston:
The sparse shift-invert Lanczos kept a DENSE n*n mass matrix and did an O(n^2) dense M*x matvec per iteration (hundreds per analysis), plus ~114MB of mass copies on a nf=2160 frame. Mass is element-consistent (sparse) everywhere:

- mass_matrix.rs: assemble_mass_matrix_2d/3d_sparse -> CscMatrix from the same element routines as the dense path (identical values per entry).
- lanczos.rs: SparseShiftInvertOp borrows &CscMatrix; M*x via sym_mat_vec (O(nnz)). Dense ops kept for small-n / sigma!=0 / constraint fallbacks.
- modal/buckling/harmonic 3D sparse paths wired to sparse mass (constraint reduction stays dense, unchanged). harmonic_modal_from_eigen shares an m_times callback so the sparse path also drops an O(p*n^2) modal-mass loop.

Parity: golden eigenvalues worst relative deviation 2.5e-15 (frame, nf=300), 2.2e-16 (shell), 0.0 (buckling 3D); op-level CSC-vs-dense < 1e-12; suite 5925 green; clippy clean.
Bench (40-story frame, nf=2160, 8 modes): e2e modal 210.3 -> 26.1 ms (8.0x); M*x matvec 3.24 ms -> 13 us (250x); mass memory ~114MB -> ~0.15MB.


Batuis:
Summary for Diego: this one is independent of the #70 → #71 → #72 stack, but it collides with it on files, so it cannot be sequenced independently.

**What it changes**

Eigen and mass-matrix numerical paths — a CSC mass matrix on the sparse eigen routes, touching `solver/{buckling,harmonic,mass_matrix,modal}.rs` and `linalg/lanczos.rs`.

**Where it collides**

`solver/modal.rs` is also modified by the #70–#72 stack and by #69. `linalg/lanczos.rs` is also modified by #74. So although the PR itself has no dependency, resolving these files in isolation means resolving them again later.

**Suggested sequence**

Settle the merge order for #69–#74 first, rebase this onto whatever will actually precede it, then have the engine owner validate against that base. Validating against `main` right now says very little, because `main` is not the base this will land on.

I verified topology and file scope only. The numerical behaviour was not audited.

PR #74, opened 2026-07-24
perf(engine): iteration hotfixes — time-history, arc-length, diagrams…

diegokingston:
…, sparse Cholesky

Four verified hotspots, each arithmetic-identical (no expectation changes):

- time_integration: hoist the time-invariant influence vector r and M*r out of the step loop (was a dense O(n^2) matvec per step per direction).
- arc_length: factor the tangent once per iteration and solve both RHS with it (was two full factorizations per corrector iteration).
- postprocess/diagrams: sort point loads once per element-force record instead of clone+sort on every compute_diagram_value_at call — the innermost function of envelopes and beam stations. Hot call sites converted (envelopes, beam stations, section stress, plastic scans); influence.rs stays on the wrapper (different ef per call).
- sparse_chol: NumericCholesky holds Rc<SymbolicCholesky> instead of an O(nnz_L) deep copy per numeric call (hit every P-Delta/nonlinear iteration). Rc over borrow to avoid self-referential SparseShiftInvertOp.

Full suite green (all 22 targets, incl. the 5925-test validation target); clippy clean (no new warnings). numeric_cholesky bench neutral at plate sizes (win grows with nnz_L and call frequency). solve_displacement_control keeps the same double-solve pattern — noted as a trivial follow-up.


Batuis:
Diego: this is iterative-solver and postprocessing work, so it needs an engine owner, and it has the same ordering problem as #73.

**Scope**

Sixteen files under `engine/`, covering arc-length, P-Delta, plastic, time-integration, and the diagram / beam-station / section-stress postprocessing.

**Overlaps with the other open engine PRs**

`solver/linear.rs` and `postprocess/combinations.rs` with the #70–#72 stack; `linalg/lanczos.rs` with #73; `solver/linear.rs` and `solver/time_integration.rs` with #69.

**What validation should look like**

Because these are iterative paths, validation needs to cover more than final values: convergence behaviour, iteration counts, failure modes, and the numerical regression suite should all be compared with the existing implementation. Any intentional differences should be documented and justified. That should run against the base this actually merges onto, which means merge order and rebase come first.

Not solver-audited from my side; this is a scope and coordination note.

PR #75, opened 2026-07-24
perf(web): batch 2D combos into one WASM call + cheaper worker/pool init

diegokingston:
- solver-service: solveCombinations2D now makes one solveMultiCase2D call (cases + combos + envelope) instead of per-case validate+analyze+solve round trips plus per-combo full-results re-serialization. Load mapping is byte-identical to the per-case path (shared buildSolverLoads2D). Models with constraints/connectors/sliding joints or duplicate case names keep the legacy per-case path (engine multi-case ignores those), and any multi-case failure falls back to reproduce the exact legacy error string.
- 3D parallel path: drop the JSON.parse(serializeInput3D()) double round trip (new input3DToWireObject). Batch combine+envelope needs an engine export — left a TODO(perf) with the exact shape needed.
- Worker pool: fetch the wasm binary once (shared cached promise), compile once on the main thread, structured-clone the WebAssembly.Module to each worker (was: fetch twice + byte copy + compile per worker).
- wasm build: drop --no-opt so wasm-pack runs wasm-opt — binary 4.22MB ->
  3.20MB (-24%); full suite passes with the optimized binary.

Tests: new solver-service-combos-2d.test.ts (7 tests — superposition math, batch/per-case parity, constraint + duplicate-name guards, error strings). Full web suite 2675 passed / 7 skipped; production build green.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -7,7 +7,7 @@
     "dev": "vite",
     "build": "vite build",
     "preview": "vite preview",
-    "wasm": "cd ../engine && wasm-pack build --target web --out-dir ../web/src/lib/wasm",
+    "wasm": "cd ../engine && wasm-pack build --target web --out-dir ../web/src/lib/wasm --no-opt",
     "test": "vitest run",
     "test:watch": "vitest"
   },
--- a/web/package.json
+++ b/web/package.json
@@ -7,7 +7,7 @@
     "dev": "vite",
     "build": "vite build",
     "preview": "vite preview",
-    "wasm": "cd ../engine && wasm-pack build --target web --out-dir ../web/src/lib/wasm --no-opt",
+    "wasm": "cd ../engine && wasm-pack build --target web --out-dir ../web/src/lib/wasm",
     "test": "vitest run",
     "test:watch": "vitest"
   },
--- a/web/src/lib/engine/__tests__/solver-service-combos-2d.test.ts
+++ b/web/src/lib/engine/__tests__/solver-service-combos-2d.test.ts
@@ -1,0 +1,127 @@
+/**
+ * Contract tests for solveCombinations2D after its reroute through the
+ * engine's batch multi-case API (solve_multi_case_2d). Pins:
+ *  - result shape + superposition math of the batch path,
+ *  - parity with the per-case path (validateAndSolve2D per case),
+ *  - the guards that keep constraint / duplicate-name models on the
+ *    id-keyed per-case path,
+ *  - error conditions still surfacing as string returns.
+ */
+
+import { describe, it, expect, beforeAll } from 'vitest';
+import { initSolver } from '../wasm-solver';
+import { solveCombinations2D, validateAndSolve2D, type ModelData } from '../solver-service';
+
+/** Cantilever: fixed at node 1, tip loads at node 2 split across two cases. */
+function beamModel(): ModelData {
+  return {
+    nodes: new Map([
+      [1, { id: 1, x: 0, y: 0 } as any],
+      [2, { id: 2, x: 5, y: 0 } as any],
+    ]),
+    elements: new Map([
+      [1, { id: 1, type: 'frame', nodeI: 1, nodeJ: 2, materialId: 1, sectionId: 1 } as any],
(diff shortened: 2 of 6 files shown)


Batuis:
Short version for Diego: the batching is a good change and I would keep it. What blocks it is that the new batched path skips the 2D model validation the old path ran, so a model that should be refused can come back with numbers instead.

Audited commit `b7379a5944378df634a5bdcf48ea5f995cd78a23`, base `main`, merge-base `c967f0393b8b516a1294b8876b736f515bf54e31`. `main` has since moved to `c13a9842`, so a merge will be needed regardless.

**What checks out**

I compared the batched route against a faithful replica of the legacy per-case route (the same code this PR moved into `solveCombinations2DFallback`) on a cantilever, a portal frame with distributed / linearly varying / nodal / point-on-element / thermal loads, with and without self-weight, plus single-case, partial and total missing case references, and a zero-load case.

Per-case results were **bit-identical** — max absolute difference `0.000e+0`, exact JSON identity, with and without self-weight. Combination and envelope values differed by roughly one ULP (max relative `3.0e-16`, max absolute `4.5e-13`), which is accumulation-order noise from combining inside the engine instead of per-combo on the JS side. The name-keyed mapping back to ids preserved key identity and insertion order in every case, including combos referencing missing case ids and cases carrying no loads.

**The blocker**

In `web/src/lib/engine/solver-service.ts`, the batched block (around lines 834–877) goes straight from `buildSolverInput2D` to `solveMultiCase2D`. `validateAndSolve2D` runs about 21 model-level guards before it solves (roughly lines 241–564). The fallback route inherits those guards because it calls `validateAndSolve2D` per case; the batched route does not.

Most guards are recovered by accident, because the engine throws and the `catch` drops into the fallback — a load referencing a deleted node, and an orphaned node, both produced the exact legacy error string in testing. But two guards are 2D/3D mode policy that the engine cannot reproduce, so nothing throws:

- **Non-zero Z coordinate on any node in a 2D model.** Legacy returns "This model has nodes with Z coordinates (3D geometry). Switch to Stabileo 3D to analyze it." The batched path returns a full result bundle.
- **A `fixed3d` support in a 2D model.** Legacy returns the equivalent "3D supports … not compatible with 2D analysis" error. The batched path returns a full result bundle.

`fixed3d` is the dangerous one. The 2D engine does not understand that support type, so the restraint set it actually applies is not the one the model describes. The reactions and internal forces that come back are wrong, and nothing signals it.

**Suggested fix**

The bypassed guards are all load-independent, so they only need to run once. Extracting them into something like `validateModel2D(model)` and calling it at the top of `solveCombinations2D`, before the batch-versus-fallback decision, covers both routes. A narrower version is adding the two mode-policy conditions to the existing fallback guard near line 821.

**Smaller items**

- A structurally malformed multi-case result (`solver-service.ts:878-880`) returns a generic envelope error rather than falling back. Only *thrown* errors reach the fallback at line 901. That branch should call `solveCombinations2DFallback` too, otherwise a solvable model can appear unsolvable behind a vague message.
- None of the seven new tests can distinguish the batched path from the fallback. They would all still pass if `solve_multi_case_2d` disappeared, because the two paths are numerically equivalent — and one of them ("unstable model") demonstrably runs the fallback today. Worth asserting that the batch actually ran.
- `wasm-solver.ts:172` — `wasm.default(await getWasmBytes())` emits `using deprecated parameters for the initialization function` on every boot. The object form `{ module_or_path: … }` avoids it.
- `solver-pool.ts` caches `initPromise` with no `.catch`, so a single failed init disables the pool until reload — even though `getWasmBytes` deliberately clears its own cache so a retry is possible.

**Engine-side, not for this PR**

Calling `solve_multi_case_2d` 20 times with byte-identical input returned 9 distinct results; the legacy route returned 1 distinct result across 10 calls. Two symptoms: `elementForces[].distributedLoads` comes back in one of two orders, and values drift by about one ULP. The magnitude is structurally irrelevant, but it breaks reproducibility for reports and for any exact-comparison test. This lives inside the engine and should go to the engine owner — please do not work around it in the web layer.

PR #76, opened 2026-07-24
perf(engine,web): JsValue boundary on the 8 hot WASM exports

diegokingston:
solve_2d/3d, solve_multi_case_2d/3d, combine_results_2d/3d and compute_envelope_2d/3d no longer cross as JSON text (stringify -> copy -> from_str -> to_string -> copy -> parse). They now take/return JsValue via serde-wasm-bindgen (maps as plain objects); export names unchanged, ~70 cold exports untouched.

- Numeric keys were a non-issue: id maps are HashMap<String,_> and results are Vec-based, verified with round-trip checks (bench-wasm-boundary.mjs).
- Semantic gap fixed: JSON.stringify used to coerce NaN->null (rejected by serde_json); serde-wasm-bindgen passes NaN through and the 3D solver propagates it. assertFiniteWire() guard in the 8 wrappers + worker.
- Worker messages now carry structured-cloned objects instead of ~1MB JSON strings (no main-thread stringify/parse on the parallel combo path).

Measured (bench script, medians): solve total is noise-bound (boundary is ~2% of solve); combine+envelope per call 1.15x-1.37x faster, and worker combos drop the main-thread JSON work entirely.

cargo test 5925 green; web suite 2675 green (43/43 boundary-robustness); build + tsc clean.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -13,26 +13,41 @@ pub fn init() {
     console_error_panic_hook::set_once();
 }
 
-/// Solve 2D linear static analysis. JSON in → JSON out.
-#[wasm_bindgen]
-pub fn solve_2d(json: &str) -> Result<String, JsValue> {
-    let input: types::SolverInput = serde_json::from_str(json)
-        .map_err(|e| JsValue::from_str(&format!("Parse error: {}", e)))?;
-    let results = solver::linear::solve_2d(&input)
-        .map_err(|e| JsValue::from_str(&e))?;
-    serde_json::to_string(&results)
-        .map_err(|e| JsValue::from_str(&format!("Serialize error: {}", e)))
+/// Serialize a result to a JsValue with maps as plain JS objects, matching the
+/// shape `JSON.parse(serde_json::to_string(...))` produced on the JS side.
+/// Used by the hot-path exports to skip the JSON text round trip.
+fn to_js_value<T: serde::Serialize>(value: &T) -> Result<JsValue, JsValue> {
+    serde::Serialize::serialize(
+        value,
+        &serde_wasm_bindgen::Serializer::new().serialize_maps_as_objects(true),
+    )
+    .map_err(|e| JsValue::from_str(&format!("Serialize error: {}", e)))
 }
 
-/// Solve 3D linear static analysis. JSON in → JSON out.
+/// Deserialize a hot-path input passed straight from JS (plain objects, no
+/// JSON text). Id-keyed maps are `HashMap<String, _>` on the Rust side, so JS
+/// object keys (already strings) round-trip natively.
+fn from_js_value<T: serde::de::DeserializeOwned>(value: JsValue) -> Result<T, JsValue> {
+    serde_wasm_bindgen::from_value(value)
+        .map_err(|e| JsValue::from_str(&format!("Parse error: {}", e)))
+}
+
+/// Solve 2D linear static analysis. JsValue in → JsValue out (hot path).
 #[wasm_bindgen]
-pub fn solve_3d(json: &str) -> Result<String, JsValue> {
-    let input: types::SolverInput3D = serde_json::from_str(json)
-        .map_err(|e| JsValue::from_str(&format!("Parse error: {}", e)))?;
+pub fn solve_2d(input: JsValue) -> Result<JsValue, JsValue> {
+    let input: types::SolverInput = from_js_value(input)?;
+    let results = solver::linear::solve_2d(&input)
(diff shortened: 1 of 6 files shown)


Batuis:
For Diego, up front: the conversion itself looks value-preserving, but two things stop it here. It breaks the worker pool for constrained models, and when I ran the PR's own benchmark with real before/after artifacts it showed no speedup.

Audited commit `223a6ba461331d5488c6444b63f662c230141dc1`, reviewed incrementally against #75 (`b7379a5944…`), which is its base. Engine changes are confined to `engine/src/lib.rs`.

**Boundary correctness**

I built both boundaries from identical solver code — this PR only changes `lib.rs` — and compared all eight exports through the public interface. `solve_2d`, `combine_results_2d` and `solve_multi_case_2d` came back byte-identical. `combine_results_3d`, `compute_envelope_2d` and `compute_envelope_3d` differed by about one ULP. `solve_3d` and `solve_multi_case_3d` showed larger differences, but those are pre-existing engine nondeterminism rather than anything this PR did: the *same* glue returns 10 distinct results in 10 calls, on the old JSON boundary as well, mostly reordered `structuredDiagnostics`. `serialize_maps_as_objects(true)` does the right thing — results contain no Map, BigInt, `undefined` or typed arrays, and survive structured clone.

**The blocker**

The 3D wire object carries `constraints` straight through from the store, and on a model that has constraints that is a live Svelte `$state` proxy array. `postMessage` cannot structured-clone a proxy. Instrumenting `Worker.prototype.postMessage` in the running app against a constrained PRO model shows it failing on every case message:

```
key="constraints" type=array
Failed to execute 'structuredClone' … [object Array] could not be cloned
```

The pool then logs `Parallel solve failed, falling back to sequential` and solves on the main thread. Results stay correct, so it is invisible outside DevTools — but the parallelism is gone. The path is `solver-service.ts:181` (`constraints: model.constraints ?? []`) → `wasm-solver.ts:372` → `solver-pool.ts:131`. Models without constraints are unaffected, because `?? []` produces a fresh array.

The correction is to snapshot / de-proxy the reactive state before the wire object is built or posted — `$state.snapshot()`, which the project's own proxy-safety convention already prescribes for anything that gets serialized — plus a test that genuinely structured-clones a wire object built from a constrained model. The current tests cannot catch this: they use an in-process fake worker, so clone is never exercised.

**The performance premise**

The PR ships `web/scripts/bench-wasm-boundary.mjs`, which expects a "before" artifact at `/tmp/wasm-old`. I built one from #75's engine and ran the author's own harness:

```
 300 elem  OLD 87.5 ms (ser 0.1 / wasm 86.6 / parse 0.9)    NEW 88.7 ms   0.99×
           combine4   OLD 9.7 ms                            NEW 10.0 ms   0.97×
1000 elem  OLD 660.0 ms (ser 0.3 / wasm 657.2 / parse 2.6)  NEW 659.7 ms  1.00×
           combine4   OLD 29.5 ms                           NEW 29.5 ms   1.00×
```

The JSON round trip was never the bottleneck for `solve_3d` — 2.9 ms out of 660. Where it did dominate (combine at 300 elements, 5.8 ms of 9.7), serde-wasm-bindgen's per-property interop costs roughly what V8's native JSON codec did. The one genuine gain I found was `compute_envelope_3d` at 1.10× raw, which drops to 1.02× as shipped because `assertFiniteWire` eats the difference. Worth either narrowing this to the payload-dominated exports or reconsidering it — as it stands it adds boundary risk without a measured payoff. If your hardware disagrees, that changes the conclusion, and it is worth checking before dropping the work.

**Smaller items**

- Non-finite rejection now lives only in the JS `assertFiniteWire` guard. Calling the new glue directly with `fx: NaN` reaches the solver and returns "Singular stiffness matrix — structure is a mechanism", where the old boundary rejected it at parse. A check inside `from_js_value` would keep the invariant at the boundary itself.
- `lib.rs:31` formats errors as `Parse error: {e}`, but serde-wasm-bindgen's Display already begins with `Error:`, so messages read `Parse error: Error: …`. Worse, an `undefined` field now yields `invalid type: unit value, expected f64` with no field name, where the old path said ``missing field `my` ``.
- `assertFiniteWire` re-walks arrays that came out of the engine (`wasm-solver.ts:699`, `:707`); those values are already known finite.
- There are no contract tests for the eight converted exports. The bench has good assertions but is not part of `npm test`.

Inherited from #75 and unchanged here: the 2D validation bypass, the malformed-multi-case fallback gap, the wasm-bindgen deprecation warning, and the batch-path test coverage gap. Those belong on #75, not this PR.

PR #79, opened 2026-07-27
[16] Adopt CIRSOC 2025 regulations, with edition provenance that cannot lie

Batuis:
# [16] Adopt CIRSOC 2025 regulations, with edition provenance that cannot lie

**Base:** `pr/15-result-invalidation-trust-repair` (`8cbf6c53`)
**Head:** `pr/16-cirsoc-2025-load-codes` (`b796ecae`) — frozen
**Commits:** 12 · **Diffstat:** 203 files, +169 502 / −390 (143 of those files are the converted official texts)

Second of five stacked PRs. Depends on `[15]`. `[17]`, `[18]` and `[19]` depend on this.

---

## Why

CIRSOC 201-2005 is not the edition in force. **Resolución 11/2026** (RESOL-2026-11-APN-SOP#MEC), published in the Boletín Oficial on 21 Jan 2026, enacted CIRSOC **101-2025, 102-2025, 200-2024 and 201-2025** with effect from 22 Jan 2026. Stabileo was implementing the superseded edition, and its wind layer was a 2005 partial that produced a velocity pressure with no pressure coefficients at all — which is not a wind load per the regulation.

Adopting a new edition is the easy half. The hard half is **provenance**: making it structurally impossible for a result to cite a rule it did not apply. That is what most of this branch is about, and it is why it ends by *removing* a capability rather than adding one.

## User workflow

Analysis ▾ → Design → **Project regulations**. One selector per *role* — basis, loads, wind, seismic, concrete, steel, masonry, timber — not one per code. A project binds one adapter per role it needs; whether that adapter is Argentine, European or American is not the product's business.

1. Bind **basis** and **loads** (CIRSOC 101-2025), **wind** (CIRSOC 102-2025), **concrete** (CIRSOC 201-2025). New projects get these already applied.
2. Changing a **load-affecting** role stages a *pending* change and sends you to Loads, where the before/after preview lives. Design asks; Loads decides.
3. In Loads → **Auto-generate from code**: pick the Table 4.1 occupancy, the live-load reduction, the wind parameters; preview the delta; apply.
4. Applying invalidates downstream revisions precisely — loads → analysis → design — so a stale result can never be read as current.
5. Changing a **design-only** role (concrete) applies in place, because the forces do not move.

## Architecture: roles, not codes

`lib/codes/roles.ts` holds a `ROLE_CATALOG` of `RoleOption`s. `basis` and `loads` are deliberately separate: EN 1990 supplies combinations while EN 1991-1-1 supplies magnitudes, and a project may legitimately mix them. Every option carries `nameKey` + `edition`, and the rendered label interpolates the edition — so the two CIRSOC 201 adapters can never both read "CIRSOC 201", which was a shipped defect. A module-scope invariant throws at import if any two options in a role render identically.

`lib/codes/regulation.ts` is the machine-readable half of `docs/codes/CIRSOC/CAPABILITY-INDEX.md`: regulation identity, edition, legal instrument, and `ClauseRef`. `assertSameEdition` **throws** rather than warns — a certificate citing two editions of one regulation is not a cosmetic defect, it is an incorrect engineering document.

`lib/codes/capability.ts` splits every capability into five independent facets — `verify`, `generate`, `coordinate`, `document`, `gate`. The defect this replaces: one boolean meant both "the app can rate a curtailment you gave it" and "the app can produce one", and the UI read the second. `gate` is what stops "we could not check punching shear" from becoming a floor marked complete.

## Coverage

| Regulation | Edition | Status |
|---|---|---|
| CIRSOC 101 | 2025 | Implemented — Table 4.1 occupancies, §4.7.2 live-load reduction, §2.3.2 combinations |
| CIRSOC 102 | 2025 | Implemented — velocity pressure **and** pressure coefficients, exposure, topography, enclosure |
| CIRSOC 201 | 2025 | Implemented — flexure, shear, axial-flexure, biaxial, slenderness, ties, spacing, Table 9.7.6.2.2 |
| CIRSOC 201 | 2005 | **Reserved, not selectable** — see below |
| CIRSOC 200 | 2024 | Named and cited only; aggregate limits taken from CIRSOC 201 §26.4 |
| INPRES-CIRSOC 103 | I 2018 / II 2005 | Reachable through the seismic **role**; detailing lands in `[19]` |
| CIRSOC 301 | 2018 | Present in the surface, honestly `UNSUPPORTED` |

The nine official PDFs all carry a real embedded text layer, so **no OCR was used**. The conversion lands as tracked Markdown with clause anchors; the ~56 MB of PDFs are gitignored with SHA-256 recorded so any conversion can be re-verified. CIRSOC 201-2025 alone is 657 pages and 2 151 anchored clauses.

## Table 9.7.6.2.2 — one authoritative evaluator (`2f72c427`)

The maximum spacing of shear-reinforcement legs was written by hand in three places. All three carried the same four defects:

| # | Defect | Severity |
|---|---|---|
| 1 | `if (VsReq <= 0) return min(0,8·d, 300 mm)` | **Unconservative — false pass** |
| 2 | Row-1 cap 300 mm where the table prints 400 mm | Conservative, still wrong |
| 3 | Only the along-length column implemented | **Missing mandatory requirement** |
| 4 | Row threshold `1/3` where the table prints `0,33` | Selected row 1 for row-2 demands |

**Defect 1 is the one that mattered.** The table is indexed on "V_s **requerido**", so a required V_s of zero is squarely inside row 1 and the row-1 limit applies. `0,8·d` appears nowhere in the table, in §9.7.6.2 or in its commentary. On the measured qa-8 beam (f'c 30, b 300, d 512, V_c 140,2 kN) it **permitted 300 mm where the table permits 256 mm** — a spacing the regulation forbids could be certified.

**Defect 3** is a clause-body requirement, not a refinement: §9.7.6.2.2 reads *"La separación máxima de ramas de la armadura de corte a lo largo de la longitud del elemento **y a través del ancho** del elemento debe cumplir con la Tabla 9.7.6.2.2."*

The table was read off the **rendered page 209 of 656** (expediente IF-2025-136960277-APN-DNGPO#MOP), not the flattened Markdown, because the `400 mm` and `200 mm` cells **span all four columns of their row**. The conversion loses column spans and makes them look like footnotes to the along-length column. Modelled as a per-**row** cap.

There is now exactly one implementation, `lib/codes/cirsoc201/transverse-spacing.ts`, consumed by candidate generation, verification, memos, UI and reports. The three copies were deleted, not patched. Prestressed members return a structured *unsupported* result — the table's 3h/4, 3h/2, 3h/8, 3h/4 columns are real provisions and prestressing is not implemented; row 1's 3h/2 exceeds d for ordinary sections, so silently reusing the non-prestressed column would be a check that was never performed.

Across-width limits are on leg-to-leg spacing, centre to centre, over `bw − 2·cover − d_stirrup`; `requiredLegs = max(2, 1 + ceil(span / acrossMax))`. The clause states no leg-to-face limit and none is invented. `legOffsetsAcross` is the single source of leg coordinates so generator, verifier and drawing cannot disagree.

## Why CIRSOC 201-2005 is unavailable (`b796ecae`)

An earlier revision of this branch applied the 2025 table to 2005 projects and *declared* the substitution, arguing it could not cause a false pass because the 2025 table is at least as restrictive in every cell (2005 caps at 600 mm vs 400 mm, and 2005 carries no across-width provision at all).

**The safety argument was true and the provenance argument was wrong.** Conservatism is not provenance. A certificate stamped "CIRSOC 201-2005" whose governing stirrup spacing came from the 2025 table names a rule it did not apply, and the reviewing engineer cannot see the difference.

The 2005 text **is not supplied**: `docs/codes/CIRSOC/` holds no 201-2005 document, no converted directory exists, and `REGULATIONS` records `textAvailable: false`. So the rule cannot be written, and this repository's own stated policy applies — `SOURCES.md`: data absent from the supplied text *"is marked unsupported in code rather than guessed"*. Same precedent as withdrawing an assumed INPRES-CIRSOC 103-II 2021 edition in favour of the edition actually supplied.

```ts
type EditionAvailability = 'AVAILABLE' | 'UNAVAILABLE_SOURCE' | 'UNSUPPORTED';
```

Three states, not a boolean, because the **remedies differ**. `UNAVAILABLE_SOURCE` means the app would implement this edition but its official text is not supplied — supplying it is what unblocks you. `UNSUPPORTED` means not implemented at all. Telling someone who owns the 2005 text that it is "not supported" implies never.

Enforced in the **catalogue layer, naming no regulation**:

- `optionsForRole` filters to `AVAILABLE`. The regulations panel dropped 2005 with **zero UI change**.
- `allOptionsForRole` serves surfaces that *explain* rather than select. The label-uniqueness invariant moved onto it — a reserved option colliding with a live one would only bite the day it is switched on.
- `bindRole` **throws** on a non-available option. Every caller is either a control fed from `optionsForRole` (which cannot offer one) or a loader (which must decide what to do instead).
- The adapter refuses at its **capability gate**, so the outcome is `UNSUPPORTED`. Letting the search run and come back empty would report `SEARCH_EXHAUSTED`, which asserts the code-permitted envelope was explored and found wanting — a claim about the *regulation*. Only `UNSUPPORTED` is true.
- `beamShear` and `ties` lose `verify`+`generate` under 2005 and **keep `gate`**, with a stated limitation.
- An unknown or absent edition fails **closed** rather than defaulting to the implemented one.
- The panel lists reserved editions read-only under "Editions not available for design" with the reason. Visible, not silently absent.

**Blast radius, stated plainly:** transverse reinforcement is required in every beam, so **no beam reaches VERIFIED under the 2005 adapter**. Measured on the production path — all eight QA-fixture members return `UNSUPPORTED` with zero candidates, zero verifier calls and `envelopeExhausted: false`. 2025 is untouched and still designs the same eight to `VERIFIED`.

### Migration / compatibility

**No migration workflow is provided, and none is needed.** The product has no active legacy users, so there is no corpus of 2005 projects to protect. A stored project naming 2005 is handled without data loss and without a false claim:

- the role-shaped loader **unsets** the binding and emits `regulations.migration.editionWithdrawn`;
- the v1 CIRSOC-specific shape binds the **edition in force** and emits the same notice.

Results stored under 2005 were produced by rules the app no longer applies, so re-running the design is the only honest outcome. Nothing silently reinterprets an old result as current.

### Re-enabling 2005 later

Supply the text → convert it under `docs/codes/CIRSOC/markdown/` → add an adapter, clause map, capabilities and tests → flip the catalogue entry to `AVAILABLE`. **No UI change.** A gate reads the source of `optionsForRole` and asserts it names no regulation or edition, so this stays true.

## Exactly what is supported

**Supported:** CIRSOC 101-2025 loads and combinations · CIRSOC 102-2025 wind incl. pressure coefficients · CIRSOC 201-2025 beam flexure/shear, column axial-flexure/biaxial/slenderness/ties, bar spacing §25.2.x, Table 9.7.6.2.2 both columns · role-based regulation stack with revision invalidation · clause provenance on every implemented rule · EN/ES.

**Not supported, and gated:** CIRSOC 201-2005 (text not supplied) · torsion · curtailment *generation* (verification only) · joint shear · slabs, walls, foundations, diaphragms (`[18]`) · seismic analysis and detailing (`[19]`) · prestressing · non-rectangular sections · CIRSOC 301, Eurocode, ACI (named, honestly unsupported).

## Tests and measured results at `a133b062`

| Gate | Result |
|---|---|
| Vitest | **3 325 passed**, 5 skipped, 159 files |
| Playwright (real-user) | **64 passed** — `regulations` 24, `single-regulation-source` 10 (EN+ES), plus `rc-design`, `rc-design-visual`, `engine-i18n` |
| Production build | OK, 13.90 s |
| i18n gates | 17 passed — engine purity, EN/ES parity, placeholder parity, no key-as-text, no duplicate keys |
| E2E hook gating | 5 passed — a normal production build contains **no** test-hook code |
| `git diff --check` | clean |
| Solver no-change | **0** files matching `^engine/`, `*.rs`, `Cargo.*`, `*.wasm` |
| Signatures | 13/13 `G`, author **and** committer `Bauti <syngoviano@gmail.com>` |

Notable suites: `transverse-spacing` **59** boundary tests (immediately below / exactly at / immediately above every row threshold, d = 400 mm and d = 800 mm in both rows and both directions, every leg-count gap, V_s = 0, negative and low positive V_s, narrow and wide webs, the reduced effective depth, the prestressed case, plus a 192-case sweep proving `requiredLegs` always yields a satisfying spacing) · `edition-availability` **14** · `edition-provenance` **13** · `roles-revisions` **53** · `row2-crossties` **10**.

## One selector, one resolved code

The regulations panel this PR introduces was not the only place a regulation could be chosen.
A second dropdown sat beside *Compute demands / Run code check / Design*, carrying the older
adapter catalogue, and it listed **CIRSOC 201 twice** — both adapters hardcoded the same
`name`, so 2025 and the unavailable 2005 rendered identically.

It was not cosmetic. That dropdown wrote `verificationStore.activeCodeId`, and that state —
not the project's `concrete` binding — is what the code check, the design run and the batch
dialog actually read. Three sources could disagree at once:

| surface | edition it used | adapter it used |
|---|---|---|
| code check, design, certificates, documents | toolbar | toolbar |
| detailing | **Project Regulations** | **toolbar** |
| `design-run` fallback | `model.codeSettings.concreteEdition` (legacy) | toolbar |

A member could therefore be verified under one edition's clauses and detailed under the
other's, and the certificate would name whichever the toolbar happened to hold.

**The repair.** `regulationsStore.concreteDesignCode()` is now the only resolver. It returns
the bound adapter id, or `null` when the `concrete` role is unbound, bound to something
`UNSUPPORTED`, or incomplete — and `concreteDesignProblem()` says which, as a structured
`EngineMessage`. `activeCodeId` and `setActiveCode` are deleted, as is the legacy
`codeSettings.concreteEdition` bypass. The toolbar keeps a **read-out** (`active-concrete-code`)
that is not a control, and the design commands are **gated, never silently defaulted**: with no
usable concrete regulation they disable and show the named reason plus a jump to the panel.

The duplicate label is fixed at its source — the adapter's `name` is now edition-qualified
(`CIRSOC 201-2025`), so no two registered adapters can render the same string.

**Roles stay independent.** This is one selector *per role*, not one regulation for the
project. `basis`, `loads`, `wind`, `seismic`, `concrete`, `steel`, `masonry` and `timber` are
each bound separately and a mixed stack remains representable; a test asserts it.

`single-regulation-source.test.ts` (15) holds the shape open by reading the sources: no
`<select>` and no `listDesignCodes` in the toolbar, no `activeCodeId` state, no consumer
reading `getDesignCode(verificationStore.activeCodeId)`, every consumer reading
`concreteDesignCode()`, and no `codeSettings.concreteEdition` bypass. A source gate rather
than a behavioural one, because the defect was a *second surface existing*, which behaviour
tests cannot see. `single-regulation-source.spec.ts` (10, EN and ES) is the visible half.

## Manual QA

Dev server: `cd web && npm run dev` → **http://127.0.0.1:4000/** (not `localhost`; port 4000 is reserved for this app).

1. **2005 is gone but explained.** Analysis ▾ → Design → *Project regulations*. Concrete selector lists only **CIRSOC 201 (2025)**. Expand **"Editions not available for design"** → CIRSOC 201 (2005), badge *Official text not supplied*, with the reason and the no-substitution statement. *(e2e R1b/R1c/R1d)*
2. **The design toolbar has no selector.** Analysis ▾ → Design. Beside the commands there is
   a **read-out** — *Concrete: CIRSOC 201 (2025)* — with no dropdown and nothing to click.
   Unbind concrete in *Project regulations*: the commands disable and name the reason with a
   jump back to the panel; they never quietly fall back to a default. *(e2e S1–S5)*
3. **Every role has a selector.** basis, loads, wind, seismic, concrete, steel, masonry, timber. Seismic reaches INPRES-CIRSOC 103 through the role, not a hardcoded tab. CIRSOC 301 is listed and says it is not implemented. *(R2)*
4. **Load-code change stages, does not apply.** Switch **wind** to CIRSOC 102 (2005) → pending banner → *Review in Loads* → preview shows before/after → Apply. Cancel instead and the applied binding returns unchanged. *(R5a–R5c)*
5. **Design-only change applies in place.** Re-bind **concrete** → no pending banner, state *Applied*, and no solve is triggered. *(R5d, R15)*
6. **Regulation-backed loads.** Load `rc-design-qa-8` → Conditions ▾ → Loads → *Auto-generate from code*. The dialog names the regulations. Pick a Table 4.1 occupancy; try one that Table 4.1 cross-references — it is **refused, not invented**. The live-load reduction is a real control. *(R6–R13)*
7. **Aggregate size lives in Materials.** The regulations panel shows it as a *requirement* with a jump button; it is edited in Materials, because it is a property of the concrete mixture, not of the regulation.
8. **Language.** Switch EN/ES; the regulations panel, the loads dialog and every clause note must be complete in both.

## Risks and reviewer warnings

- **The 2005 retirement is a product decision, not a defect fix.** If any real project needs 2005, this branch blocks it by design. That was the explicit instruction; it is reversible only by supplying the text.
- **Table 9.7.6.2.2 re-baselines design outcomes.** Fixing defect 1 changes which members pass. Published incorrect results are not compatibility requirements, but the change is real and visible in `[17]`'s qa-8 outcomes.
- **`0,33` vs `1/3`** shifts the row boundary by 1,01 %. Intentional and per the printed table, but it moves a threshold.
- **`DEFAULT_SPACING_RULE` changed the default edition** from 2005 to 2025. Any caller that omitted `spacingRule` now gets the edition in force. That is the fix, but it is a behaviour change for un-migrated callers.
- **The same provenance question remains open for other 2005-cited rules.** `spacing.ts` §7.6.1/§7.6.2/§7.6.3 keeps its edition branch (2005 genuinely differs — no aggregate term), and its numbers are now unreachable from any project because `cirsoc-2005` cannot be bound. The branch is retained deliberately so re-enabling an edition does not mean rebuilding the mechanism; a gate asserts it still branches.
- **169 k insertions look alarming** — 143 files of that is the converted official text corpus. The reviewable code diff is ~50 files under `web/src/lib`.

## Reviewer guide — highest-risk files first

| Risk | File | Commit |
|---|---|---|
| **Highest** | `lib/codes/cirsoc201/transverse-spacing.ts` — the rule everything else consumes | `2f72c427` |
| **Highest** | `lib/engine/station-design-forces.ts` — the authoritative verifier; read the shear-check block and `DEFAULT_SPACING_RULE` | `2f72c427` |
| High | `lib/codes/roles.ts` — `bindRole` refusal, `optionsForRole` filter, both migration paths | `b796ecae` |
| High | `lib/engine/design/adapters/cirsoc201-adapter.ts` — the `unsupported()` capability gate | `b796ecae` |
| High | `lib/engine/design/candidate-enumerate-beam.ts` — the generator must start at `requiredLegs` or the search reports a false `SEARCH_EXHAUSTED` | `2f72c427` |
| Medium | `lib/codes/regulation.ts` — `EditionAvailability`, `assertSameEdition` | `b796ecae` |
| Medium | `lib/engine/loads/*`, `lib/codes/cirsoc101/*`, `lib/codes/cirsoc102/*` | `f2285e8a` |
| Medium | `components/pro/design/ProjectRegulationsPanel.svelte` | `a2507cc3`, `b796ecae` |

Read `2f72c427` and `b796ecae` first; the earlier ten commits are the adoption groundwork and are lower-risk.

## Solver no-change proof

```
git diff --name-only pr/15-result-invalidation-trust-repair pr/16-cirsoc-2025-load-codes \
  | grep -Ei '^engine/|\.rs$|Cargo\.(toml|lock)$|\.wasm$' | wc -l
→ 0
```

No Rust, no Cargo manifest, no generated WASM, nothing under `engine/`. All work is in `web/src` and `docs/`.

## Stacked dependency

Merge after `[15]`. `[17]` consumes `transverse-spacing.ts`, the capability matrix and the role model; `[18]` and `[19]` consume the clause/provenance layer. Do not merge out of order — `[17]`'s stirrup zones will not compile without this branch's evaluator.


diegokingston:
### Code review

Found 3 issues:

1. `regulations` and `revisions` are declared on `StructureModel` and `ModelSnapshot`, but `snapshot()` builds an explicit object literal that omits both and `restore()` never assigns them. Since [`file.ts`](https://github.com/lambdaclass/stabileo/blob/44d7c9a5caba2bbb7e096ead4544dae4c23d627a/web/src/lib/store/file.ts#L101-L103) saves via `modelStore.snapshot()` and loads via `modelStore.restore(data.snapshot)`, and `history.svelte.ts` undo/redo uses the same pair, every save/open, autosave, tab restore and undo silently drops the whole role stack and the revision vector — `regulationsStore` then falls back to `defaultRegulations()`/`emptyRevisions()`, losing the bindings and the staleness stamps that gate results. `migrateRegulations` is also never called outside tests, so the documented 2005-project migration notice never fires. (bug — the fields are new in this PR, so this is not pre-existing)

https://github.com/lambdaclass/stabileo/blob/44d7c9a5caba2bbb7e096ead4544dae4c23d627a/web/src/lib/store/model.svelte.ts#L942-L950

2. `cirsoc101-2005-basis`, `cirsoc101-2005-loads` and `cirsoc102-2005` carry no `availability` field, so `availabilityOf` defaults them to `AVAILABLE` and `roleUsable` accepts `IMPLEMENTED_PROVISIONAL` — they are selectable in the regulations panel and usable. But `regulation.ts` records `textAvailable: false` for both editions (the same status as concrete 2005, which is correctly marked `UNAVAILABLE_SOURCE`), and `cirsoc102/wind.ts`, `cirsoc101/combinations.ts` and `cirsoc101/live-loads.ts` have no edition parameter and no 2005 branch — every clause ref is hardcoded `'2025'`. Selecting "CIRSOC 102 (2005)" therefore yields a project stamped 2005 whose wind loads are computed by, and cite, the 2025 rules. That is the defect [`optionsForRole`'s own docstring](https://github.com/lambdaclass/stabileo/blob/44d7c9a5caba2bbb7e096ead4544dae4c23d627a/web/src/lib/codes/roles.ts#L251-L255) says the model exists to prevent. The guard test is scoped so it cannot catch this: it asserts `expect(o.role === 'concrete').toBe(false)`, which passes vacuously for basis/loads/wind.

https://github.com/lambdaclass/stabileo/blob/44d7c9a5caba2bbb7e096ead4544dae4c23d627a/web/src/lib/codes/roles.ts#L165-L172

3. Wind force is not distributed by height. `wind.ts` evaluates the windward wall at two heights specifically [“so the caller can distribute”](https://github.com/lambdaclass/stabileo/blob/44d7c9a5caba2bbb7e096ead4544dae4c23d627a/web/src/lib/codes/cirsoc102/wind.ts#L426-L430), ordered `[min(5, h), h]`. The caller takes `.find(...)`, which returns the first — the weakest, z = 5 m — sample, computes `net` once outside the per-level loop, and applies it unchanged to every level; the mean-roof-height value is never read. Since `Kz = 2.41·(max(z,5)/zg)^(2/α)` rises monotonically with height, this under-predicts wind at every level above 5 m (roughly 20% at z = 15 m, 30% at z = 30 m for 45 m/s / exposure C). The `applyMinimumWindLoad` floor does not mask it for ordinary multi-storey buildings, so the error is unconservative and flows into member sizing.

https://github.com/lambdaclass/stabileo/blob/44d7c9a5caba2bbb7e096ead4544dae4c23d627a/web/src/lib/engine/loads/load-plan.ts#L423-L436

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #100, opened 2026-08-04
perf(web): combine/envelope guard only on factors, not on solver output

diegokingston:
## What

`assertFiniteWire` recursively walked the **full payload** on every combine/envelope call — multi-MB per-case result trees, per combination. The results are solver output (produced after the input-side guard already rejected non-finite model data), so the walk was pure cost. It now guards only the factors (user data, cheap). The full input guard on solve inputs is unchanged.

## Measured (1000-element 3D model, 4 cases per combine)

| Payload | Guard walk | Combine call | Guard share | Saved per 10-combo batch |
|---|---|---|---|---|
| 1000 elem / 3.2 MB | 3.44 ms | 14.18 ms | **24%** | ~34 ms |
| 5000 elem / 14.8 MB | 19.9 ms | 69.2 ms | **29%** | ~199 ms |

1 file, +10/−8. Web suite: 3063 passed, 0 failed.

### DIFF
--- a/web/src/lib/engine/wasm-solver.ts
+++ b/web/src/lib/engine/wasm-solver.ts
@@ -672,9 +672,11 @@ export function combineResults(
     .filter(f => perCase.has(f.caseId))
     .map(f => ({ caseId: f.caseId, results: perCase.get(f.caseId)! }));
   if (cases.length === 0) return null;
-  const payload = { factors, cases };
-  assertFiniteWire(payload);
-  return wasmCombineResults2d(payload);
+  // Guard the FACTORS only (user data, cheap). The per-case results are solver
+  // output — produced after the input-side guard already rejected non-finite
+  // model data — so re-walking multi-MB result trees per combo was pure cost.
+  assertFiniteWire(factors);
+  return wasmCombineResults2d({ factors, cases });
 }
 
 /** Combine 3D results with factors via WASM. JsValue in/out — no JSON round trip. */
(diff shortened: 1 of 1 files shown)


diegokingston:
### Code review

Found 1 issue:

1. The premise "the per-case results are solver output — produced after the input-side guard already rejected non-finite model data" holds only for the 2D *unconstrained* path, so removing the guard leaves NaN undetected at every layer rather than removing a redundant check.

`solve_2d`/`solve_3d` delegate to `engine/src/solver/constraints.rs` for any model carrying a rigid link, equal-DOF or diaphragm constraint, and that file contains no `is_nan`/`is_infinite` check anywhere — `solve_constrained_2d` and `solve_constrained_3d` take `cholesky_solve`'s `Some(u)` straight through. `cholesky_solve` only rejects a pivot `<= 1e-15`; it never inspects the solved values, which is exactly why the 2D unconstrained path added an explicit `has_nan_inf` check *after* calling it:

https://github.com/lambdaclass/stabileo/blob/fa097b03871f856b316190116880d1ee0e8daf47/engine/src/solver/linear.rs#L586-L592

The constrained paths and 3D `solve_loads_dense` never got that check, so they can return `Ok(results)` containing NaN from entirely finite input. `wasm-solver.ts` does not check solver *output* either — the only `Number.isFinite` in the file is inside `assertFiniteWire` itself — and `engine/src/postprocess/combinations.rs` has no finiteness handling at all. With the guard gone there is now nothing on any layer.

The failure mode is worse than the throw it replaces. In `compute_envelope` the running max is seeded from `results.first()` and advanced with `if d.ux.abs() > displacements[i].ux.abs()`; under IEEE-754 every comparison against NaN is false, so a NaN in the first case is never overwritten and permanently poisons that field, while a NaN in a later case is silently dropped — the worst case vanishing from the envelope that exists to catch it. Neither raises anything. Previously this threw `Non-finite number (NaN) at input[0].displacements[3].ux`, naming the exact path.

https://github.com/lambdaclass/stabileo/blob/fa097b03871f856b316190116880d1ee0e8daf47/web/src/lib/engine/wasm-solver.ts#L697-L712

Worth separating the two halves: dropping it from `computeEnvelope`/`computeEnvelope3D` follows Batuis's #76 review directly ("`assertFiniteWire` re-walks arrays that came out of the engine (`wasm-solver.ts:699`, `:707`); those values are already known finite"). Narrowing `combineResults`/`combineResults3D` to `factors` extends that reasoning to the `cases` argument, which that review did not cover — and `computeEnvelope` is never called on raw `solve()` output in production, only on `allComboResults` from `combineResults` itself ([`solver-service.ts:1075`](https://github.com/lambdaclass/stabileo/blob/fa097b03871f856b316190116880d1ee0e8daf47/web/src/lib/engine/solver-service.ts#L1070-L1076), `:1860`, `:1960`), so the chain now runs through combine's own arithmetic.

The same #76 review also flagged the durable fix and it was never landed: "Non-finite rejection now lives only in the JS `assertFiniteWire` guard... A check inside `from_js_value` would keep the invariant at the boundary itself." Suggest landing that, or a `has_nan_inf` check in `solve_constrained_2d`/`solve_constrained_3d` and `solve_loads_dense` matching the existing 2D pattern, together with this removal — that catches the blow-up where it originates with the existing "Singular stiffness matrix — structure is a mechanism" error, and keeps the perf win.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #125, opened 2026-08-10
PR20 — PRO UI and design workflow

Batuis:
> [!NOTE]
> ## Integración con la base — preparada, sin verificar
>
> La base `pr/19-rc-cad-constructibility` avanzó 37 commits. La integración está **hecha**: los 9
> conflictos fueron resueltos a mano en `pr20-base-integration-wip` (`dd0dcf22`), cada uno
> documentado en el mensaje del commit. **Ningún gate post-integración fue ejecutado, y no se
> afirma ninguno.**
>
> Una versión anterior de este bloque decía que la base publicaba TypeScript contra bindings WASM
> nunca regenerados, y pedía trabajo a @diegokingston. **Eso era incorrecto y queda retirado.**
> `web/src/lib/wasm/` está en `.gitignore` y no tiene archivos versionados: los bindings son un
> artefacto local por worktree, generado con `npm run wasm`. Los dos errores de tipo venían de un
> build local viejo en ese worktree, no de la rama. Si typecheck falla por `analyze_section_*_field`,
> correr `npm run wasm` ahí — no es un defecto de la base ni del solver.
>
> Detalle en `docs/handoffs/pr20-readiness.md` §5b.

> [!WARNING]
> **Estado de PR20 · draft**
>
> - Alcance funcional **cerrado y aceptado**.
> - Gates **verdes contra la base con la que se desarrolló** (`d585b7ed`): unit 5989 · build
>   tests 14 · production build limpio · paridad de locales 60 · smoke 37/37 · suite completa
>   **402 passed**, 1 failed (screenshot conocido de 1 px), 4 skipped, 1 did not run.
>   **Esos números no describen el árbol integrado.**
> - **QA manual: pendiente.** Checklist en `docs/handoffs/pr20-qa-manual.md`.
> - **No mergeable**: `CONFLICTING` + el bloqueo de arriba.
> - Listo para **revisión** de @diegokingston.

---

> ### This PR is PR20.
>
> It was opened as #125 against `main`. It has since been **retargeted onto `pr/19-rc-cad-constructibility`**
> and is now the twentieth entry in the RC chain — `main → PR19 (#90) → PR20 (this PR)`.
> GitHub numbers are permanent, so it keeps the number 125; semantically it is **PR20 — PRO UI and
> design workflow**. No substitute PR was opened, and none should be: this is the one.
>
> Base: `pr/19-rc-cad-constructibility` (PR #90). PR19 brings the RC CAD constructibility review —
> the 3D viewer, the large workspace, the `VERIFIED / PROVISIONAL / FAILED / UNSUPPORTED` states,
> torsion warnings, the biaxial proposal, plans, autosave and restoration. This PR carries the PRO
> shell and the advanced-analysis repairs on top of it.

**This is an interface PR for PRO mode.** 12 commits of PRO work, all under `web/` — **nothing in `engine/` is touched**, and nothing needed to be. That matters for the handoff: everything below was fixed on the JavaScript side of the WASM boundary.

Draft — for Bauti's review, and for the agent taking this forward.

---

## Part 1 — the shell (the point of the PR)

PRO is a **pipeline**, not a loop:

```
model → conditions → solve → results → demands → code check → design → detailing → report
```

with state that persists between stages and can go stale. Basic's shell answers a question PRO does not ask, so the tabs are justified here in a way they were rejected there.

**What is built:**

- **Two-level ribbon.** Quick tools (Project / Save / Undo / Redo, then Select▾ / Pan) on the left of the tab row, at the same height and hierarchy as the tabs; then four stage tabs with badges; then command groups with labels underneath. The whole group row collapses.
- **A Project view that exists.** PRO had none: saving was a shortcut, opening was a file input owned by another component, exporting sat at the bottom of an unrelated tab, and "Examples" opened a floating gallery over the canvas. Now file / new model / export / share, with the example gallery **inline in the panel** and `?` help on the DXF and IFC importers.
- **One thing at a time in the right panel.** Results (8 `<details>` → a section strip with counts and a fallback when the current one is empty), Loads (3 sections → a selector), Advanced (17 `<details>` → a chip picker). The panel has a heading that names what it shows, an error count, and a visible hint to reopen it once closed.
- **Results in the ribbon**, as in Basic 3D — N / My / Vz / Mz / Vy / T, with a scale slider on every diagram.
- **The palette migrated by role.** A nearest-colour rule sent dark blues to green, so the rule became: anything dark is a surface whatever its tint; anything washed out is text; only a saturated mid-tone may claim a semantic token, and only if its hue belongs to that token's family. Known intents are pinned. The turquoise that did three jobs (a value, a button fill, a heading) is split by the job each declaration does.
- **Design badge** on the tab, from the verification summary.

**Left open, on purpose** (see "For whoever picks this up"): the Design stage is barely touched, at Bauti's request.

---

## Part 2 — what the audit turned up along the way

Two findings that are worth reading before touching this area. Both were fixed here; both are corrections to the **web** layer.

### A. Every colour map was dead in the default render mode

`syncColorMap3D` walked `ctx.elementGroups`, which is **not the element registry**. In wireframe render mode — the default — a member without a hinge is drawn by the batched `LineSegments2` and gets no `Group` at all (`needsGroup` in `scene-sync.ts`), so on a model without hinges that map is **empty**. The loop painted nothing while still setting `colorMapApplied = true`.

That silently killed **axial member colour, the stress heatmap and the CIRSOC verification colours** — the reported symptom was only the first of the three. Verification ratio labels had the same defect.

Fixed by iterating the model and treating the `Group` as optional decoration. Regression test: `viewport3d/__tests__/colour-map-wireframe.test.ts` runs a colour map with **no groups at all** and reads the colour off the batched mesh.

### B. Thirteen of the seventeen advanced analyses had never run

The panel builds a JSON payload per analysis and hands it to a WASM export that deserialises into a named Rust struct. Thirteen sent a payload the engine has never accepted, failing at the deserialiser. Nobody saw it because every `catch` read `e.message ?? 'Error'` — and a WASM rejection has no `.message` — so the panel reported the single word "Error", thirteen times over.

| Analysis | Was sending | The struct wants |
|---|---|---|
| Harmonic | `fMin/fMax/nPoints/dampingXi/direction` | `frequencies[]`, `dampingRatio`, `responseNodeId`, `responseDof` |
| SSI | `direction:"Y"`, curve `softClay`, `gamma` | axis index, `py_soft_clay`, `gamma_eff` + `eps_50` |
| Staged | `{solver, stages}`, no name | `StagedInput3D` is **flat**; stages need `name` / `elementsAdded` / `elementsRemoved` |
| Creep | one `concrete` block, `time` | EC2 parameters **per material**, steps keyed on `tDays` |
| Cable | a bare solver input | `{solver, densities}` |
| Model reduction | DOF indices | boundary **nodes** |
| Multi-case | `caseIds` | the load vectors per case, plus combinations |
| Section analyser | `{shape:'rect', b, h}` | `polygons` of `[y, z]` vertices |
| Constrained | `{nodeA, nodeB}` + a `method` | a tagged `rigidLink` constraint |
| Imperfections | `type` + `amplitude` | a notional-load block |
| Arc-length / disp. control | a 3D model | a 2D `SolverInput` |

Several result read-outs were reading fields no result type has, so an analysis that *did* succeed printed a blank line: Winkler claimed a convergence it never reports, pushover looked for a load factor instead of its collapse factor, the section analyser for `area` instead of `a`. Winkler and the constrained solve return `AnalysisResults3D` directly, so their results never reached the viewport either.

**Arc-length, displacement control, cable and model reduction are 2D-only in the engine** and a PRO model is spatial, so all four collapsed on the way in whatever the user did. They are **removed from PRO's picker** — a chip that can only ever be grey is worse than no chip. Their wrappers stay exported and their contracts stay tested for whoever surfaces them in a 2D workspace.

**Result: 13/13 of the analyses PRO offers now run and report on a solved model, with 0 page errors.**

Controls that could not do anything are gone: the custom p-y curve had no fields to fill it, and the creep per-step load factor had no counterpart in the engine.

New: `engine/__tests__/advanced-wire-contracts.test.ts` — 16 tests calling each wrapper with the payload the panel builds, against the real WASM engine. The deserialiser *is* the thing under test, so a payload that stops parsing is a test failure rather than the word "Error".

---

## For whoever picks this up

1. **You will not need to touch the solver.** None of the above did. Every defect was on the JS side of the boundary — payload construction, result field names, which registry a loop walks. If something in PRO looks like an engine bug, audit the wrapper and the payload first; that is where all thirteen lived.
2. **The audit is not finished.** It covered the Advanced tab end to end and the 3D colour maps. The Design stage was deliberately left alone, and the other tabs were only checked for *opening cleanly*, not for producing correct output. The same class of defect — a payload or a result field written from memory rather than from the struct — is plausible anywhere the web layer talks to WASM. `advanced-wire-contracts.test.ts` is the pattern to copy.
3. **Open design decision, deliberately not taken:** the Design stage's commands live on a stage-level command bar rather than in the panel that owns them. Either move the commands into the panel, or make the stage bar the rule for every stage. Note that `cmd-generate-detailing` lives on the design toolbar, not on the detailing panel — a stage-level selector that ignores this breaks detailing (tried and reverted).
4. **Do not repaint element colours from a loop over `elementGroups`.** See finding A.

---

---

## Part 3 — the conversion to PR20, and what it surfaced

This PR was retargeted from `main` onto `pr/19-rc-cad-constructibility` and PR19 was merged in
(`c49bd367`). No substitute PR was opened; #125 is PR20.

### The five conflicts, and why each was resolved the way it was

None was taken as "ours" or "theirs".

| File | PR19 brought | PR125 brought | Kept |
|---|---|---|---|
| `playwright.config.ts` | the fuller account of the shared-preview-port hazard | the detail that the adopted build also lacks `VITE_E2E=1` | both, in one comment |
| `OutcomeBadge.svelte` | a `provisional` state | the `--st-*` migration | migration + the new state |
| `DesignToolbar.svelte` | the `summary-count-provisional` chip | the `--st-*` migration | migration + the chip |
| `ProDesignTab.svelte` | `overflow: auto` + a 14rem table floor | a 9rem floor in `DesignTable` | both floors |
| `App.svelte` | the restore banner, and two additions to `pro-bar` | the inline prompt, and the ribbon | the inline prompt and the ribbon — see below |

**The provisional violet stays a literal.** `#a066d3` is load-bearing in six places, and its
authority is `three/rebar-scene.ts`, which hands a numeric hex to a Three.js material and cannot
read a custom property; `run-summary-reported.test.ts` asserts the toolbar agrees with it by
value. Tokenising the two CSS sites alone would split one definition in two. `--st-provisional`
is owed — see `docs/handoffs/pr20-ui-and-workflow-plan.md` §5.6, which predicted exactly this.

**Both branches had independently found the same design-table defect** from opposite ends: PR19
that the tab could not scroll at 1280×720 so the table collapsed to 0 px, PR125 that the table
could shrink under its own sticky header so a click on row 1 hit "select all". The second ribbon
row makes PR19's the more necessary of the two.

**The autosave banner was not re-mounted.** PR124, which this branch carries, had just moved that
prompt inline beside the tabs because a full-width banner in that slot pushes the application
down on every load. PR19 re-added the banner on its own branch. Mounting both would mean two live
Restore buttons and the layout defect back. PR19's contribution to it — the timestamp, and the
warning that what you are about to restore is not your newest save — is grafted onto the inline
prompt, carried by the `⚠` and the sentence rather than by the amber. The two specs that reached
for `.autosave-banner` now reach for `data-testid="autosave-prompt"`.

**PR19's `pro-bar` gives way to the ribbon,** and both things PR19 had added to it survive in the
Project view: `ProProjectFileActions`' entry points, and the `Plano DXF` button firing the same
`stabileo-import-dxf` event under the same label, now with a `?`.

### One thing added beyond the merge

`fix(a11y): the 3-D workspace stops lying about being modal` (`fa84fc9b`). The overlay declared
`role="dialog" aria-modal="true"` and managed no focus at all: Tab walked into the page ARIA had
just called non-existent, opening left focus on the button underneath the overlay, and Escape
dropped focus to `<body>`. `pr20-ui-and-workflow-plan.md` §5.2 lists this first as the only
accessibility item that makes the feature unusable rather than degraded. Mechanism in
`lib/utils/dialog-focus.ts`, covered by `e2e/rebar-workspace-focus.spec.ts`.

The rest of §5.2 is untouched and still open: no keyboard route to the conflict markers, no
landmark or heading structure in the rail, no text alternative for the canvas.

### The blocker this conversion surfaced — PR19 against main, not against this PR

Six unit tests fail on this branch. **All six fail identically on `pr/19` merged with `main`
with none of this PR present**, so they are a PR19 ↔ main collision that the retarget exposed,
not a merge defect here:

| Test | Expected | Got |
|---|---|---|
| `beam-reinforcement-audit` — accounts for every beam | 117 `PROVISIONAL` | 5 |
| `beam-reinforcement-audit` — states the shared cause once | 117 | 5 |
| `beam-reinforcement-audit` — proposes only for genuinely biaxial beams | ratio > 1 | 0.2438 |
| `top-steel-projections` — shows it in the status table | > 50 proposals | 0 |
| `design-families` — counts processed / designed / refused | > 50 | 5 |
| `scene-completeness` — reports beams honestly | > 50 named proposals | 5 |

PR19 changes **nothing** under `engine/`. Main moved 75 files there since PR19's merge base —
`cirsoc201_check.rs`, `ec2_check.rs`, `check_ledger.rs` and the rest of #113–#119, the pass that
corrected five design-check formulas, added missing limit states, stopped reporting
non-evaluable checks as passing, and derived defaults from the standard. PR19's detailing
expectations were calibrated before that landed.

Whether 5 or 117 is the right number is a design-check question, not a UI one. It belongs to
whoever owns PR19 and the detailing engine, and is deliberately untouched here.

---

## Part 4 — the PRO workflow, driven end to end

Four reports, and they turned out to share one shape: a control that was technically correct
and told the user nothing true.

**The settings button did nothing.** It toggled its state and mounted its panel — into
`.app-body`, which is `position: relative`, so the panel's `top: 100%` meant "one whole app-body
down", off the bottom of the window. On PR19 its positioned ancestor was `<nav class="pro-bar">`
and it landed under the bar; the ribbon replaced that nav and nobody noticed the anchor had gone
with it. It is now anchored to the gear, with `aria-expanded` / `aria-controls`, a heading, a
close button, Escape, close-on-click-outside and focus in-and-back. The spec measures the
panel's box against the viewport, because `toBeVisible()` passes on a panel that is merely below
the fold.

**`Ver modelo 3D` is on the Design command row**, beside `Design all` — disabled with a reason
while there is nothing to draw, carrying the assembly count when there is. The button in the
detailing disclosure stays beside the exports. Both call `openRebar3D`, so they are one
operation; a spec opens the workspace from each and compares the scene census. That also lifted
the engineer's name into `detailing-author.svelte.ts`, because with two callers a
component-local `$state` meant the same revision got signed two different ways.

**The warning no longer greets people for having an empty workspace.** `checkModel` returns
three errors for an empty model and the chip rendered on `count > 0`, so PRO opened with "⚠ 3"
before the user had done anything. `diagnostics-warning.svelte.ts` now requires a model, an
armed session, a surviving error-severity diagnostic, and a set the user has not dismissed.
Dismissal is per diagnostic SIGNATURE for the session: it lapses when the diagnostics change and
does not survive a reload. **It can never change what the design does** — the commands and the
pre-solve gate call `checkModel` themselves, and there is a test that says so. The chip moved
from the panel header, where it rendered on every tab, to the right of the Design row.

**The Project section has a gutter and a card per decision.** `.pro-content` is `padding: 0` and
that tab added none, so its buttons sat against the panel border. It now states what is open,
what the autosave is doing and where the storage is — as STATUS. There is still exactly one
offer to restore a project, the inline prompt beside the tabs, and a spec asserts this panel
does not grow a second one.

One bug of my own, found by the new unit test rather than in review: a `$derived` declared in a
module-level factory has no effect root to invalidate it, so the first read cached forever. It
looks correct inside a component and is wrong everywhere else. Those values are plain getters
now, memoised on the model counts.

### Gates for this pass

| Gate | Result |
|---|---|
| `npm run typecheck` | 479 / 490 baseline, no new errors |
| `npm run build` | clean |
| `npm run test` | **5 929 passed, 6 failed** — the same six PR19 ↔ main failures, unchanged |
| Playwright `@smoke`, `E2E_PORT=4293` | **130 passed, 0 failed**, 4 skipped |
| `pro-workflow-shell.spec.ts` (new) | 13 passed |
| `diagnostics-warning.test.ts` (new) | 9 passed |
| Rust / Cargo / WASM / solver files changed | **0** |

## Gates

Re-measured on the integrated branch, on its OWN engine build: `web/src/lib/wasm/` in this
worktree was an August 9 artefact, which predates #127's AISC repair. Every number below is
from a fresh `npm run wasm`.

| Gate | Result |
|---|---|
| `npm run typecheck` | 479 against the 490 baseline — **no new errors** |
| `npm run build` (production) | clean |
| `npm run test` — `build` pass | 8 passed |
| `npm run test` — `unit` pass | **5 920 passed, 6 failed** — all six are the PR19 ↔ main collision above, all six reproduce without this PR |
| Playwright `@smoke`, `E2E_PORT=4293` | **112 passed**, 4 skipped, 0 failed |
| `e2e/rebar-workspace-focus.spec.ts` (new) | 5 passed — trap, initial focus, restore on Escape and on close, and the trap surviving a rail fold |
| Playwright `project-restore` + `rebar-toggles`, incl. `@slow` | **12 passed** — restore, the eight rail switches on the 7-storey building, plans, reconciliation |
| `engine/` files changed by this PR | **0** |

`E2E_PORT` matters: `reuseExistingServer` is off whenever it is set, so the suite gets its own
preview server instead of adopting a sibling worktree's — the failure both branches had
independently documented in `playwright.config.ts`.

**Not ready for review yet.** The six unit failures are a real blocker, they just are not this
PR's to clear. §5.2's remaining accessibility items are open. `--st-provisional` is still owed,
and the seven RC components PR125's colour sweep never saw still carry 74 literal hexes between
them — `pr20-ui-and-workflow-plan.md` §5.6.1 called that the largest silent integration gap, and
it is still there.


diegokingston (commented):
Read it at `44cd26d9`. **Scope, first:** 22 270 lines across 136 files is more than one honest pass, so I went where an error would be quiet rather than loud — the advanced-analysis wire contracts, `rcCheckability` and the census, and `verification-service`. I did **not** read the PRO shell components, the ribbon, the tab layouts, the e2e specs, the i18n, or the third-party notices. Call it a tenth.

## The description no longer describes the PR

Worth fixing before anyone else reads it, because every claim in the header block is now false:

| the header says | reality at `44cd26d9` |
|---|---|
| `draft` | not a draft |
| `No mergeable: CONFLICTING` | `MERGEABLE` |
| base `pr/19-rc-cad-constructibility` | base is `main` (PR19 merged) |
| "Listo para revisión de @diegokingston" | zero reviews recorded |

Also stale: the gate numbers are stamped "contra la base con la que se desarrolló (`d585b7ed`)" with an explicit warning that they do not describe the integrated tree. CI on the current head is green so far — `lint`, `test`, `suite (1)`, `suite (2)` all pass, `e2e` and `web` still running — so the numbers can simply be refreshed.

## The wire-contract audit is the best thing here

Thirteen of seventeen advanced analyses were sending payloads the engine never accepted, and two more — contact and time history — parsed **and ran wrong**: contact as a plain linear solve, time history with zero ground motion, because serde dropped the unknown fields silently. That is a genuinely hard class of bug to find, and the write-up in `advanced-wire-contracts.test.ts` is precise about which field each one got wrong.

I ran the suite: **109 passed**. The contact and time-history tests do what the header claims — they assert the input reaches the solver rather than merely parsing, which is the only way to catch the silent pair.

## The finding: the mechanism that made two of them silent is untouched

`engine/src/types/input.rs` has **78 input structs and exactly one `#[serde(deny_unknown_fields)]`**.

So the seventeen payloads are fixed, and the reason two of them failed *silently instead of loudly* is still in place for the other seventy-seven. The next time a payload field is renamed, mistyped, or dropped in a refactor, the same thing happens again: the analysis runs, returns plausible numbers, and nothing anywhere says the input was ignored.

This is not a criticism of what the PR did — its own scope note says `engine/` is deliberately untouched, and that was the right call for an interface PR. It is that the fix is at the call sites and the hazard is at the boundary. Seventeen call sites are now correct; the boundary will accept the eighteenth mistake just as quietly.

Adding `deny_unknown_fields` to the analysis input structs is a Rust change of a few lines per struct, and it converts this entire class from "silently wrong" to "parse error" — which the panel already knows how to show, since that is what the other thirteen were doing.

Related, and acknowledged in the file itself: the contract tests are **payload shapes hand-copied from the panel**, with nothing enforcing the match. That is honest, and it is also the only thing standing between seventeen analyses and a silent regression. With `deny_unknown_fields` at the boundary, the hand-copying stops being load-bearing.

## `rcCheckability` is well built, with one mislabel

The design is right and I want to say so plainly: one named predicate, used by both the loop that checks and the layer that explains why nothing was checked, with a census computed from the *same* predicate. "CIRSOC 201-2025 verified no member in this model" being true, unhelpful and indistinguishable from a bug is exactly the right thing to have gone after.

The mislabel: `if (!fc || fc > CONCRETE_FY_CEILING) return 'notConcrete'`. `!fc` is also true for a material whose strength is simply not filled in yet, so an **incomplete concrete** material is reported to the user as "not concrete". It is counted and explained, which is far better than being dropped — but the reason given is wrong, and "you have 4 non-concrete members" sends someone looking at the wrong thing. A `noStrength` case would separate "this is steel" from "I cannot tell yet".

## Interaction with #135, worth planning before either lands

Both PRs independently define `CONCRETE_FY_CEILING = 80` and both answer "is this concrete", at different granularities:

- here: `rcCheckability` → `checkable | noSection | noMaterial | notConcrete | noRectangle`, with a census;
- #135: `materialFamilyOf` → `concrete | steel | timber | masonry | aluminium | unknown`, with a *basis* attached (`declaredGrade` vs `inferredFromFy`).

#135's handoff already says `auto-verify.ts` should import the constant from `material-family.ts` once this lands. Two notes on that: #135's version distinguishes "declared by a catalogued grade" from "guessed from magnitude", which is strictly more information and is what the caveat text needs; and #135's `buildAllMemberContexts` **silently omits** non-concrete members where this PR counts and explains them. This PR's handling is the better half of that pair — worth keeping when they are reconciled rather than letting the newer file win by default.

## Smaller

- The `titleKey` addition in `verification-service.ts` is exactly right: `title` stays untouched because it is the `{#each}` key at the render site, `titleKey` is additive, and the memo BODIES are deliberately left alone because rewriting them means rewriting authority modules. That reasoning is worth copying elsewhere.

## Verdict

I would not block on anything I read. The wire-contract work alone justifies the PR. Before it lands I would refresh the header — it currently tells a reviewer the PR is a conflicted draft, which will cost someone a wasted hour — and I would open a follow-up for `deny_unknown_fields` rather than let the silent-payload class stay open behind seventeen hand-maintained tests.

Given what I did not read, this is not an approval of the whole PR; it is a report on the tenth of it where a mistake would not announce itself.

PR #132, opened 2026-08-11
basic: section analysis, a multi-code materials database, and the UI to reach them

Batuis:
## What this is

It started as an audit of Basic mode's advanced analyses and the test coverage
they were missing. It grew into the section analysis, the materials database and
the panel layout that audit showed were needed — 34 commits, and worth reading
in the order below rather than as a diff.

## 1. The original finding: analyses with no numerical validation

P-Delta, buckling, modal, plastic collapse, spectral and moving loads had **no
numerical validation anywhere**. The tests that mentioned them checked plumbing:
that the ✕ clears the right store slot, that results invalidate on edit. All
necessary, and none would notice a wrong number. Now validated against
closed-form solutions, which also surfaced a real defect: the per-element moving
load envelope missed the span maximum.

## 2. Section analysis

Normal and shear stress over the real polygon, principal stresses and Mohr's
circle, the centroid and shear centre derived step by step, the central core, a
draggable eccentric load point, and a stress map painted from the exact stress
plane.

Torsion is presented as what it is — three different theories over three
different wall topologies, which disagree where they overlap:

- **Cauchy** for circular sections, exact.
- **Bredt** for closed thin walls, and on a stocky tube it reads about 5% LOW
  against Cauchy — approximate in the unconservative direction.
- **Saint-Venant** as the general theory, with the closed form that evaluates it
  named per shape.

Rows that do **not** apply are shown with the reason, because applying Bredt to
an open section looks reasonable on paper and is wrong by two orders of
magnitude. Warping is included: omitting it underestimates stress.

## 3. Materials and profiles, four regions

CIRSOC 301, Eurocode 3, AISC 360 and NBR 8800, with 700+ profiles across 15
families and grades audited against their published standards. Concrete and
timber on the same footing as the metals, with the modulus by code (CIRSOC/ACI
`4700√f'c`, EN `22000·((fck+8)/10)^0.3`, NBR `5600√fck` — a 42% spread at
20 MPa).

Each family records the grades it is really rolled in, per region and more than
one: a W is A992 today and was A36 for decades, and both are on drawings.
Choosing a profile brings its grade with it — but only when no steel was chosen
on purpose, so a verification of an existing structure is never overwritten. A
departure from practice is stated, never blocked.

## 4. Three languages, complete

Basic mode is at 100% in English, Spanish and Portuguese, enforced by a test
that walks the source for every key the UI asks for. The language menu offers
those three, detects the browser and falls back to English.

## 5. The panel and canvas work

The ribbon and the data tabs are two views of one selection; results live beside
the controls that choose them; editing and reading a result are mutually
exclusive, enforced as an invariant rather than as an action each entry point
has to remember.

Canvas labels are laid out against what is actually on screen — largest first,
ties broken D/L/W/E, measured against the members, the arrows and each other in
a single pass per frame.

## Defects worth calling out

Several were invisible to the tests that existed, because each piece passed its
own and none could see the seam between them:

- `stationForces3D` read `txStart`; every 3D result produces `mxStart`. The read
  was optional, so **every torsional and warping stress in the app was zero**
  while the panel displayed the torque one row above the words "no torque at
  this station". The call site had `as never` on it, which is what stopped the
  compiler from saying so.
- The stress map's gradient carried one of its two minus signs, so it painted
  the compressed flange red — the exact opposite of the σ diagram beside it.
- `--st-border` was referenced 23 times and defined nowhere. An undefined custom
  property invalidates the declaration, so those borders were silently never
  drawn.
- Circular tubes reported **half** their Jourawski shear: one-sided Q with
  two-sided b.
- Eight rolled channels panicked the mesher through a 1.7° spike where the toe
  fillet did not fit the flange tip.

## Testing

- **5528** web unit tests (from 1117).
- **217** e2e; the blocking `@smoke` suite is green. The one failure locally is
  `overlay legend`, a macOS pixel baseline that is preexisting and
  `continue-on-error` in CI.
- Typecheck baseline lowered 490 → 479.

## Not in this PR

- PRO profile families (JIS/GB/AS).
- NBR 5884 welded VS/CS/CVS profiles — blocked on the dimension tables.
- Portuguese for PRO, Education and the landing page, which have their own PRs.


diegokingston (commented):
Read it in the order you suggested. This is very good work — the audit finding is real, the modules that came out of it are written to be understood rather than just to run, and the tests mostly pin physics instead of implementation (checking Cw against published IPE/HEB tables, asserting that using the outside dimensions would under-report the torsion stress). What follows is what I found; most of it is small, and two are worth fixing before merge.

**Scope, so you know what this covers.** I read the section-analysis physics closely — `warping.ts`, `tensors.ts`, `eccentric.ts`, `torsion-flow.ts`, `stress-map.ts`, `section-stress.ts`'s core routine — and the materials database. I did **not** review `label-layout.ts`, `box-select.ts`, `draw-loads.ts`, the i18n locales, the e2e specs, or the panel components except where they consume the physics above. 108 files is more than one pass covers honestly.

---

## 1. The Saint-Venant share ignores which end restraint you picked

`warping.ts:240`. `bimoment` branches on `restraint`; `saintVenantShare` does not:

```ts
const bimoment = restraint === 'cantilever'
  ? absT * lambda * Math.tanh(ratio)
  : (absT * lambda / 2) * Math.tanh(ratio / 2);
...
const saintVenantShare = Math.max(0, Math.min(1, 1 - 1 / Math.cosh(ratio)));
```

So the two numbers the same call returns come from two different boundary-value problems. `TorsionDetails.svelte:186` renders the share as a percentage directly beside the restraint tabs at :173 — flip to **simple** and `sigmaW` changes while the percentage does not, which is visible from the UI alone.

The `simple` bimoment is the standard torsionally-simply-supported result: by symmetry each half is a cantilever of length `L/2` carrying `T/2`, giving `B = (T/2)·λ·tanh(L/2λ)` — which is exactly what you wrote, so the bimoment is right. The share consistent with that case uses `ratio/2`. For an IPE 300 (λ = 1.28 m):

| L | L/λ | share as coded | consistent with the `simple` bimoment |
|---|---|---|---|
| 1 m | 0.78 | 0.24 | 0.07 |
| 2 m | 1.56 | 0.60 | 0.24 |
| 3 m | 2.35 | 0.81 | 0.44 |
| 6 m | 4.69 | 0.98 | 0.81 |

At 2 m it over-states Saint-Venant by 146 %. The direction is the one this module's own header warns about: a too-high Saint-Venant share reads as "warping is not doing much here", on exactly the short members where it is doing nearly all of it.

No test catches it — `warping.test.ts:148` checks that `sigmaW` differs between the two cases, never the share.

## 2. The principal strains violate trace invariance

`tensors.ts:114`. `principalOf` builds `centre = sigma/2`, which encodes "the other normal component is zero". That is true for the stress tensor (σ_y = σ_z = 0 is the beam-theory assumption you set out in the header) and false for the strain tensor, where `eps_yy = eps_zz = −ν·eps_xx` — which the lines just above compute correctly.

Pure tension, σ = 100 MPa, E = 210 GPa, ν = 0.3, no shear. With no shear the strain tensor is already diagonal, so its principal values *are* its diagonal:

```
principal strain as coded : 4.7619e-4   0.0000e+0   0.0000e+0
principal strain, truth   : 4.7619e-4  -1.4286e-4  -1.4286e-4
```

The decisive part is the trace, which no rotation can change: the principal values must sum to `eps_xx + eps_yy + eps_zz`. As coded they sum to 4.76e-4 against the true 1.90e-4 — and `volumetricStrain`, in the same returned object, computes that same trace correctly. The object contradicts itself. `maxShear` comes out 23 % low for the same reason.

The comment defends the reduction on the grounds that the principal *directions* coincide by isotropy. They do; the *values* don't follow, because the reduction hard-codes a zero that only the stress state has.

Mitigating: `principalStrain` is currently read by nothing — `StressTensorDetails.svelte` renders `principalStress`, the invariants and the strain tensor in µε, never this. So no user sees a wrong number today. It is on the exported interface though, and this is the module whose header makes a point of Poisson contraction being "the part people forget".

## 3. Two smaller things in `warping.ts`

**The share's comment puts the formula at the wrong end** (:236). It says "the Saint-Venant part at the restrained end is 1 - 1/cosh(L/lambda)". At the restrained end it is identically zero — that is what warping restraint means, and the module's own header says so. From `θ' = (T/GJ)[1 − cosh(k(L−z))/cosh(kL)]`, the coded expression is `θ'` at `z = L`, the free end where the torque is applied. The number is a reasonable thing to report; the sentence explaining it is wrong.

**The channel's sectorial coordinate is not measured about the shear centre**, though :187 says it is. `peakSectorialCoordinate` returns `h0·bm/2`, with no subtraction of the shear-centre offset `e`; the flange-tip value is `h0·(b−e)/2`. For a UPN with `e ≈ 0.3–0.4b` that over-states ω, and so σ_w, by roughly half. Conservative, so not a safety issue — but this file states the direction of every other approximation it makes, and "adequate for the magnitude" undersells a 50 % margin.

## 4. The fall-through bucket calls everything closed

`warpingProperties` ends with `klass: 'closedNegligible'` for a degenerate channel (`bm <= 0`) and for every unhandled shape, including `rect` and `generic`. For a solid rectangle "negligible" is the right magnitude but "closed" is the wrong reason, and for a malformed open section it is wrong on both. In a module where the class exists precisely to carry the reason the number is what it is, this is the one place the reason is invented.

## 5. An observation, not a defect: two opposite y/z conventions

`ResolvedSection` has `y` vertical (`yMax` = top fibre) and `z` horizontal. The canonical path has them the other way round — `eccentric-tensors.test.ts:96` builds a 0.2 × 0.4 rectangle with `iy = 0.2·0.4³/12` and `zMax = 0.2`, pairing `iy` with `z`. Both conventions are internally consistent and the one place they meet is reconciled correctly, with the comment at `SectionStressPanel.svelte:487` doing the work:

> The core polygon is stored (ez, ey) — horizontal, vertical — which is the canonical point's (y, z) in that order.

I checked the mapping and it is right. I am flagging it because that comment is the only thing standing between the two, and `computeCentralCore`'s local `const Iz = rs.iy` re-labelling on top of it makes the file harder to check than it needs to be. Worth a named adapter or at least a comment at `section-stress.ts:18` saying the other half of the app reads them the other way.

---

## The materials database

Reviewed separately and the values check out — see #145, which carries the one gap I found: `byThickness` reached no production caller, so the picker offered S355 as a flat 355 MPa with nothing on screen to say it drops to 335 above 40 mm, in the unsafe direction. That PR also corrects an attribution I got wrong on the first pass and which is worth knowing here: the bands in this file are **design-code** tables, not product-standard ones — EN 1993-1-1 t.3.1, CIRSOC 301, EN 1999-1-1 t.3.2b — and never `productStandard`. The `fu` values are the tell: EN 10025-2 quotes Rm as a range (470–630 MPa for S355), never the single 490 here. Your comment at the EN block already says this; #145 makes it data so nothing downstream can display a 40 mm step beside the label "EN 10025-2" and imply that is where it came from.

The grade-versus-code separation is the right model and is the confusion worth preventing. No other findings in `structural-grades.ts`.

`torsion-flow.ts` is the best thing I read here and I have nothing to fix in it. I checked the three theories against their limits: Bredt takes `Am` from the mid-line and uses `4Am²/(perimeter/t)`; the solid rectangle gives 0.1408·a⁴ against the tabulated 0.1406 and `τ = 4.8T/a³` against the exact 4.81, and tends to the thin-strip limit correctly. Its 19 tests pin the physics, including one asserting that using the outside dimensions would under-report the stress — the classic error the module avoids.

`eccentric.ts` is correct as far as I checked it: `T = vz·armY − vy·armZ` is the right cross product, the kern limits `r²/c` are assigned to the correct sides, and `snapShearCentre` is solving a real problem (a 10 µm solver residue times 60 kN is a torque that isn't there). One caveat worth writing down: `kernLimits` and `computeCentralCore` both assume the bounding-box axes are principal, so for `L`/`invL` — where `Iyz ≠ 0`, which the canonical solve does carry — the painted core and the numbers agree with each other and not with the tilted neutral axis the stress map draws. Both are wrong together rather than disagreeing, so the marker won't turn green outside the orange region; but the region itself isn't the kern for an angle.

## Verdict

Not blocking, and I'd merge this once **1** is fixed — it is a wrong number on screen, next to the control that is supposed to change it. **2** I would fix in the same pass because it is three lines and the module is otherwise scrupulous. **3**, **4** and **5** are comment-and-labelling work you can take or leave.

Happy to push any of these to your branch if that is easier than doing them yourself — say the word.

diegokingston:
Prepared the fixes for **1**, **2** and the channel one in #152, stacked on this branch the same way #145 is — merge it in, cherry-pick it, or close it and keep just the tests, whichever suits you. Nothing in it touches anything outside `warping.ts` and `tensors.ts` and their tests.

Two things from writing them that are worth adding to the review above:

- Your `simple` bimoment is **right** — `(T/2)·λ·tanh(L/2λ)` is the torsionally-simply-supported result, and finding that is what pinned down that the share was the one that had drifted, not the bimoment.
- The channel tests anchor themselves by walking the mid-line and integrating `ω(s) = ∫ r_t ds`, then checking the integration reproduces **your** `Cw` — a completely different route to the same warping constant. It matches to 0.00 %, so that independently validates the channel `Cw` formula as well, which I had only checked against a published value before.

I left the `closedNegligible` fall-through alone: doing it properly needs a new `WarpingClass` and a label across 14 locales, and your i18n is your call rather than mine.

diegokingston:
You beat me to two of the three — and your fixes are the same ones I had written, so I dropped mine rather than leave two versions of the same change in flight. #152 is rebased onto 2d3e8ec and now carries only the channel finding.

For the record, since we arrived at them independently:

- `saintVenantShare` on the half-span parameter — same. Your `the simple case is the cantilever solution on the half-span` covers it, so my share tests went too.
- `principalOf` taking both normal components — same. You pass `otherNormal` as the third eigenvalue where I passed `strain.zz`; identical here because the two contractions are equal, and yours is the simpler signature. Worth a comment at some point that the third value is only `otherNormal` while `eps_y = eps_z`, so nobody wires up a state where they differ and gets a quiet wrong answer.
- `kernLimits` returning null when `iyz` is not negligible is a better answer than the one I had in mind. I was going to suggest documenting the principal-axes assumption; refusing is stronger, and the "solver noise must not trip the guard" case is the right second test.

What is left in #152 is `peakSectorialCoordinate` in `warping.ts` — a different module from the `section-teaching.ts` shear centre you fixed, and still measuring the flange tip from the web while its comment says it measures from the shear centre. 60% high on a UPN 200, straight into `sigma_w`.

The tests there might be worth keeping regardless of what you do with the fix: they anchor by integrating `ω(s) = ∫ r_t ds` along the mid-line and checking the result reproduces your `Cw` by a different route. It matches to 0.00%, so it validates the channel `Cw` formula independently of the published value.

Still untouched from the review, and still your call: the `closedNegligible` fall-through for degenerate and unhandled shapes.

diegokingston (commented):
Second pass, on `8e2240b` — after #152 and #145 landed. This round I checked your review fixes adversarially rather than re-reading the whole PR, then followed one of them into the production path. Your three physics fixes are right; following the CHS one downstream turned up something bigger than what it fixed.

## Your fixes, checked

- **`saintVenantShare` on the half-span** — correct, and `the simple case is the cantilever solution on the half-span` pins it.
- **`principalOf` taking both normal components** — correct. One note: the third eigenvalue is `otherNormal`, which holds only while `eps_y = eps_z`. True for every state this builds; worth a line saying so, or nobody will notice the day a state has them differ.
- **Channel shear centre in `section-teaching.ts`** — correct. `e = bm²hm²tf/(4·iy)` is the same result as `3bm²tf/(6bm·tf + hm·tw)`, and the `tw/2` puts the origin where the decomposition actually starts. Independently: my `warping.ts` integration in #152 puts a UPN 200's pole 26.63 mm from the web centre line, i.e. 22.4 mm clear of the outer face, which is the published value you tightened the bounds to.
- **CHS shear** — correct for a real tube. `Q = 2tR√(R²−y²)` over `b = 2t` gives `τ_max = 2V/A`, and the solid-bar pair `Q = (2/3)(R²−y²)^{3/2}` over the chord gives `4V/3A`. Both check out.

## The finding: a solid round column is analysed as a thin tube

`SectionShape` has `rect` for a solid rectangle and nothing for a solid circle. So `computeSectionProperties('circular')` and `('concrete-circular')` emit `shape: 'CHS'` with **solid** `a`, `iy` and `j` — and **no `t` at all**. `resolveSectionGeometryLegacy` then reaches

```ts
t = t ?? Math.min(b, h) * 0.05;
```

and fabricates a 5 % wall. Your `t = 0` solid-bar branch is exactly the right guard and its test pins it — but production never emits `t = 0`, it emits `t` absent, which becomes a wall. **The branch that protects the case is not the branch the case takes.**

Measured through the app's own builders, D = 500 mm, both `circular` and `concrete-circular`:

| | reported | truth |
|---|---|---|
| shear, design path (`shearStress`) | 4.000 · V/A | 1.333 · V/A |
| shear, drawn path (`computeShearFlowPaths`) | 4.000 · V/A | 1.333 · V/A |
| torsion `τ_max`, T = 50 kN·m | 5.92 MPa | 2.04 MPa |

The torsion one is the tell that this is the shape tag and not the data: `rs.j` is **already correct** — `πR⁴/2`, exactly what `section-shapes` computes. What misroutes it is `CHS` plus a wall, so all three theory rows come back `applies: true` — including **Bredt, a closed-thin-tube theory, on a solid bar**. The exact `T·R/J` is not offered at all, because there is no solid-circle shape to offer it for.

Everything here is conservative in direction. I am still calling it out because 3× conservative on a utilisation is not safe, it is unusable — a round column that is fine reads as failing, and `concrete-circular` is the ordinary RC column.

One thing to say plainly rather than leave for you to find: **for solid circles the CHS fix moved this the wrong way.** The old pairing `Q = t(R²−y²)` over `b = 2t` gave `τ = V·R²/2I = 2V/A`, 1.5× the truth; it is now `4V/A`, 3×. That is not an argument against the fix — it is right for real tubes, which is what it was aimed at — it is an argument for closing the `t = 0` gap in the same breath.

### The fix, verified end to end

Two literals: `t: 0` in the `circular` and `concrete-circular` branches of `computeSectionProperties`. I ran it:

```
as production builds it: t = 0.025
   saintVenant  applies=true   tau=5.9238
   bredt        applies=true   tau=5.6432
   cauchy       applies=true   tau=5.9238   governs
with t: 0:               t = 0
   saintVenant  applies=true   tau=2.0372
   bredt        applies=false  tau=null            ← declines, correctly
   cauchy       applies=true   tau=2.0372   governs
exact T·R/J = 2.0372 MPa
```

and both shear paths land on 1.3333 · V/A exactly. So the one change fixes the design shear, the drawn shear and all three torsion rows, and makes Bredt refuse a solid bar instead of answering for it.

Underneath it is a type question you may want to settle instead: the union can say "solid rectangle" and cannot say "solid circle", so solid circles have to borrow a hollow tag and be told apart by a sentinel. `t: 0` is the cheap fix; a `'circle'` member is the one that stops the next person hitting this.

I can push either as a follow-up to your branch — say which.

## Still open from the first pass

The `closedNegligible` fall-through in `warpingProperties`, for a degenerate channel and for `rect`/`generic`. Needs a new `WarpingClass` and a label across 14 locales, so it stays yours.

## CI

Green on `8e2240b` after both merges: `test`, `suite (1)`, `suite (2)`, `e2e` and `web` all pass. The `@smoke` `commercial-grade` failure that `2d3e8ec` introduced is resolved by your `561ffe0b`.

Worth noting as process rather than code: `2d3e8ec` broke a blocking `@smoke` and sat red on this branch for a day, because CI gates the PR and nothing gates a direct push to its head. Same family as the `@slow` suite never running on PRs.

PR #135, opened 2026-08-12
PR21 — Metallic structures in PRO: honest steel, and parametric generators

Batuis:
Branched from `main` (`542fc664`), **not** from PR19 (#90) or PR20 (#125). Neither of those worktrees was touched.

56 files, +10k / −30. All the content is in new files; there are **14 edits to existing files**, and only **4 of them touch work in flight, totalling 37 lines**. That split is deliberate — #125 touches 253 files and #132 touches 61, and the overlap was measured before anything was written. `docs/handoffs/pr21-integration.md` lists every one with its resolution.

## Two things, one branch

**Steel in PRO, honestly.** There is no usable metallic design authority in this app, so nothing here claims one. Members are recognised, listed and stated as undesigned; `steelCountsAsVerified` returns the literal type `false` so a caller cannot branch on a maybe.

**Parametric generators.** Trusses (5 kinds + monopitch), lattice columns and sheds, with previews and per-row section figures. These *can* be finished, and are: a generator produces geometry and asserts nothing about capacity, so a node position is a fact the app can establish and test.

## The defects it found and fixed

**A steel member was getting a full concrete `MemberContext`,** with its yield read as f'c. It was refused later — nothing was ever certified — but under `DEMAND_UNAVAILABLE` with the demands sitting right there, and counted in every concrete total on the way past. A steel hall with two concrete columns reported "2 verified of 302".

**The CIRSOC 301 table was showing a green tick.** Behind it: 769 lines with zero tests, no clause map, no place in the maturity model, `Lb = L` hardcoded, and `tw`/`tf`/`Fu` invented when the section or material does not state them. The numbers stay — an engineer who can see the assumptions is better served than by a blank — and they now arrive under a banner that lists those assumptions.

**Two mechanisms in my own generators,** found by a test that solves every generated geometry with the real solver. No topology test could see them: symmetry, counts and coordinates are all satisfied perfectly by a mechanism.

- Every truss emitted `rollerXZ` for its sliding bearing, on the reading that a "roller in XZ" rolls along the span. It is the opposite: `solver-service.ts:1402` maps it to `{ rx: false, ry: true, rz: false }` where the flags mean RESTRAINED — free in X *and in Z*. Every generated truss stood on one pin and one bearing that carried no vertical load.
- The shed defaulted its latticed column bases to pinned, and the lacing braces the column in its own plane and nothing else, so the pair could fold sideways. Default is now fixed; pinned is still allowed and is disclosed, with a test asserting the unstable configuration *is* unstable.

**The extruded 3-D view drew an I-beam for everything generated.** Measured: a channel box chord, a back-to-back angle post and a plain IPE all came back as the same fabricated 13-point I outline, because a generated section carried no `shape` and `createSectionShape`'s `default:` branch says "Default to I-shape". `Section.composition` now records what a section is made of, declaratively, and the viewport builds the real outline from the same placement table the properties came from. `shape` is emitted for a single profile only — on a compound one it would make the canonical resolver rebuild one part's geometry and silently replace the assembly's A, Iy and Iz.

None of these were solver defects. `rollerXZ` does exactly what its code says, and a planar frame on a bearing that does not hold it up is genuinely a mechanism.

## What it does not do

No metallic verification, no adapter bound, no Rust, no WASM, and no change to any concrete result — `rc-baseline-digest.test.ts` pins the 408-member frame member by member, because the existing aggregate gate asserts 386/22 and is blind to a change that moves utilizations while leaving the counts alone.

CIRSOC 301 becomes bindable and **experimental** rather than a red error: a steel project can state the code it is designed to, `roleUsable` still refuses it, and the stamp that reaches reports and drawings says the binding is experimental.

## Three safety nets

| | |
|---|---|
| `rc-baseline-digest.test.ts` | the concrete design, member by member. Fingerprint `1bd4d9c1d575b085` |
| `generated-models-solve.test.ts` | every generated geometry solves. The only net that catches a mechanism |
| `e2e/generators-steel.spec.ts` | four properties that must survive a UI rewrite, asserted on behaviour |

The UI is deliberately thin and is expected to be reworked; that last spec is what the rework has to keep green.

5564 unit · 8 build · 10 e2e · production build and svelte gate clean.

The one type error the gate reports is pre-existing on `main` — verified by stashing every file here — and #125 deletes the unused constant that causes it.


diegokingston (commented):
Read it against `08917b9f`. Big PR, so first: **what I actually reviewed.** The material-family classifier and the RC exclusion, the `rollerXZ` mechanism finding, the generator solve tests, and the steel inventory census. I did **not** read the parametric generators' geometry, the 3-D composition/extrusion work, the CIRSOC 301 capability table, the UI, or the e2e specs. Roughly a fifth of 14k lines, chosen because it is where an error would be quiet.

On what I read, the engineering is good and two of the three claimed defects check out exactly. One finding below is worth fixing before this lands.

## Verified

**The `rollerXZ` mechanism is real, and the reading is right.** `solver-service.ts:1402` returns `{ rx: false, ry: true, rz: false, … }` where the flags mean RESTRAINED — so free in X *and in Z*. A truss on that bearing has no vertical support at one end. Nothing subtle about it once you look; the reason it survived is that no topology test can see it, which the description says and which is correct.

**`steelCountsAsVerified` returning the literal type `false`** is the right shape for this. A caller cannot branch on a maybe, and it cannot be made to lie later without changing its signature.

**`materialFamilyOf` is the right consolidation.** One place, one rule, with the basis attached — `declaredGrade` versus `inferredFromFy` are different claims and separating them is what lets a surface say "guessed". Injecting the grade lookup so the module works before and after #132 is the right call for two branches in flight.

**The self-correction at `generated-models-solve.test.ts:169`** is the best thing in the diff:

> The claim was: latticed columns on pinned bases are a mechanism… The singular matrix that backed it was real. Its CAUSE was not the pins.

A wrong diagnosis, kept visible with the real cause beside it. That is worth more to the next reader than a clean-looking test would have been.

## The finding: an unfinished CONCRETE member now vanishes instead of being refused

`buildAllMemberContexts` filters on `ctx.materialFamily !== 'concrete'`. A material with no usable `fy` does not classify as concrete — it classifies as `unknown`:

```
material                        family     basis            included in RC?
C25 concrete (fy = 25)          concrete   inferredFromFy   yes
steel S355 (fy = 355)           steel      inferredFromFy   NO   ← intended
material with NO fy             unknown    noData           NO   ← not intended?
material with fy = 0            unknown    noData           NO
no material at all              unknown    noData           NO
```

A concrete member whose material is half-entered is an ordinary state while modelling, and it is exactly the member a user needs told about. Before this PR it appeared in the RC table blocked with `missingMaterial` — visible, with a reason. Now it is filtered out of the contexts entirely, so:

- it is not in the RC design table, with or without a reason;
- it is not in the metallic members list either — `steel-inventory.ts:127` lists only `family === 'steel'`;
- it is a number in `census.byFamily.unknown`, and `emptyReasonOf` only surfaces that as `allUnclassified` when **every** element is unclassified (`:208`). In the mixed case — 300 good concrete members and 2 blank ones — nothing fires.

So it degrades from "listed with a reason" to "a count on a panel about steel, that does not display in this case". The `providedSummary` denominator shrinks silently, which is the same class of thing this PR fixes on the steel side, pointing the other way.

Worth saying: the context builder itself handles this correctly — `steel-excluded-from-rc.test.ts:138` asserts `materialFamily === 'unknown'` and `blocking` containing `missingMaterial`. The blocking reason is computed and then discarded by the filter, so the test passes while the user never sees it. That test exercises `buildMemberContext`; nothing exercises `buildAllMemberContexts` with an unknown-family member.

The argument in the docstring — *"they are not members this pipeline refused; they are members it was never asked about"* — is true of steel and not of this. A concrete member with no strength recorded **is** a member the concrete pipeline should be asked about and should refuse loudly.

Narrowing the filter to exclude only families that are positively something else (`steel`, `timber`, `masonry`, `aluminium`) and keeping `unknown` in, blocked, would fix it and is a one-line change. Whether `unknown` belongs in the concrete pipeline at all is your call — but silently omitted is the one option I would rule out.

## Structural

- Base is `feat/pro-visual-system` (#125), which is **also open** and is 11 commits behind `main`. So this is a three-deep chain and neither link has landed. Not a defect; worth knowing before planning the merge.
- **CONFLICTING against its own base**, in three files only: `i18n/locales/{en,es,pt}.ts`. Everything else auto-merges, so it is a contained rebase.

## Smaller

- `CONCRETE_FY_CEILING = 80` is duplicated by #125's constant of the same value and name, with the follow-up recorded in the handoff. Fine as an interim, but it is two copies of a threshold that decides which pipeline a member enters — worth making sure that two-line follow-up actually happens rather than becoming the fifth copy the module's own docstring complains about.
- The description says "56 files, +10k" while GitHub reports 76 files, +14 124. Probably grown since it was written; worth refreshing so the numbers in it can be trusted.

diegokingston (commented):
Second pass at `08917b9f` — unchanged since my last review, so everything there still stands. This time I read a part I had explicitly skipped: the CIRSOC 301 capability declaration.

## The capability model is the right shape

Worth saying before the finding, because the design deserves it:

- every facet false, written once in the module header rather than repeated per key;
- `build()` **throws at import** if a metallic key has no limitation text — "an unsupported capability without a stated reason is an unexplained blank, which is what this model exists to prevent" is exactly right, and enforcing it at import rather than in a test means it cannot be merged around;
- `CIRSOC301_PROMOTION_KEY` names what would have to exist before any of it could be promoted, which is the part most maturity models leave implicit;
- `CIRSOC301_JS_ASSUMPTIONS` inventories what the existing checker invents (`Lb = L`, `tw`/`tf`/`Fu` when the section or material does not state them) instead of hiding it behind a green tick.

## The finding: `gate: true` is set, and nothing reads it

`build()` ends with a second loop over every metallic key:

```ts
// `gate` on every metallic capability: a model with steel in it cannot be reported
// complete because its concrete passed.
for (const key of STEEL_CAPABILITY_KEYS) {
  m[key] = { ...m[key], facets: { ...m[key].facets, gate: true } };
}
```

The only function that reads `facets.gate` is `gatingGaps` (`codes/capability.ts:155`), whose docstring says "A model with any of these in play cannot be reported as complete or constructible, regardless of how many other checks pass."

**`gatingGaps` has no production caller.** Searched `web/src` and `web/e2e`: three occurrences total — its own definition, its import in `codes/__tests__/regulation.test.ts`, and one assertion there. Nothing else.

I checked the other side too, in case something enforced it by a different route: `CIRSOC301_CAPABILITIES` is consumed once, at `store/steel.svelte.ts:119`, through `explainUnsupported(..., ['verify'])` — which produces explanatory text, not a gate. And `detailingReadiness` does not look at metallic members at all; every "steel" in that file is reinforcing steel.

So the comment states a protection that no production path enforces. A model with steel members whose concrete all passes can still be reported complete, because the flag that was supposed to stop it is read only by a test.

To be fair about authorship: `gatingGaps` and the unwired `gate` facet come from `f6b56560`, the five-facet capability commit — they predate this PR. What this PR does is become the first code to **rely** on that mechanism in a comment. The gap is inherited; the claim is new.

Two ways out, and they are different sizes: wire `gatingGaps` into whatever declares a model complete, or soften the comment to say the facet is declared for a gate that does not exist yet. I would not merge the sentence as written either way — this PR's own standard is that an unstated gap is the thing to prevent, and this is one.

## Still open from the first pass

The unfinished-concrete case: `buildAllMemberContexts` filters `materialFamily !== 'concrete'`, and a material with no `fy` classifies as `unknown`, so a concrete member whose material is half-entered is dropped from the RC table rather than refused in it. Unchanged at this head.

Worth adding a comparison I did not have then: **#125 handles the same question better.** Its `rcCheckability` returns a discriminated reason and feeds a census, so nothing disappears — it is counted and explained. When the two are reconciled after #125 lands, that is the behaviour to keep; `materialFamilyOf` has the better classification (basis attached, `declaredGrade` vs `inferredFromFy`) and `rcCheckability` has the better handling of what to do with the answer.

## Scope

Between the two passes I have read: the material-family classifier and the RC exclusion, the `rollerXZ` finding, the generator solve tests, the steel inventory, and now the capability declaration. Still unread: the parametric generators' geometry, the 3-D composition/extrusion work, the `Section.composition` design, the UI, and the e2e specs. This is a report on what I read, not an approval of 14 000 lines.

PR #151, opened 2026-08-15
basic: stress maps, a colour scale, and one rule for the highlight

Batuis:
Targets `main`. It began stacked on #132 and was retargeted once that merged;
`audit/basic-advanced-features` is now fully contained in main.

**Sections 5, 6 and 7 were absent from the first version of this description
and are about a third of the diff.** Section 5 in particular is a new
structural-analysis path — cut a 3D model with a plane and solve the result as
a 2D frame — and it is the part that most needs a reviewer's eyes.

## 1. Stress, as a command

Utilisation and Von Mises had no way in. Pressing `9` was the only route, and
the control that chose between them appeared once you were already there — so
in Basic, where the old diagram grid is hidden, they were unreachable.

They get a ribbon command of their own, last in the results row, because they
are derived FROM the internal forces rather than being one and belong to no
command that names a force. Four measures, since the section-stress evaluation
already produces all of them:

| | |
|---|---|
| **Utilisation σ/fy** | the default: 1.00 is the limit, whatever the steel and whatever the section |
| **Von Mises** | the combined stress |
| **Normal σ** | axial plus bending |
| **Shear τ** | the pair that answers "bending or shear?", which Von Mises blurs by design |

Two defects surfaced: the 2D map demanded a yield strength for Von Mises, which
does not use one, so every concrete member went unpainted while the number was
perfectly computable — and when the ratio genuinely cannot be computed the
panel now says so instead of painting nothing, which reads as "no stress
anywhere".

Shell Von Mises was offered in every 3D model because the gate was on the
analysis mode; it is on the model now. Basic has no way to make a plate, so a
measure for elements that do not exist is not a disabled option, it is a wrong
one.

## 2. Box select, and what a selection means

Supports and Loads drew a rectangle and selected nothing. Both now go through
the module the 2D viewport uses, so a distributed load is judged on the stretch
it covers and a support on its node, identically in both modes.

That module took a point as two numbers, which forced the 3D caller to flatten
every node to z = 0 — a marquee on a tower would have selected from whichever
storey happened to project onto it. It takes a point now, and a member's
station is measured along its true length rather than its projected one.

The filter became a **set**: with the multi-kind switch on, one sweep can take
nodes and supports and loads together. Off by default, and the default is the
point — with one kind armed, a click on a node carrying a support and a load
has exactly one meaning.

**A selection is what Delete removes.** Members used to bring their end nodes
along so the highlight would not look unfinished; sweeping a frame's members
and pressing Delete then took the nodes, and with them the supports and the
nodal loads standing on them. Each kind now brings only itself. Wanting both is
the multi-kind switch, where it is asked for rather than assumed.

Audited rather than sampled: all fifteen non-empty combinations × both gestures
× ten rectangles × two projections, by property — composition, isolation, and
window ⊆ crossing — so the per-kind tests that pin real numbers carry the whole
combined space. Checked against three deliberate mutations of the module; they
fail four, five and eight of the assertions, so the audit can in fact fail.

## 3. A colour scale

A gradient with no scale says where the peaks are and not how big they are; the
same red is 40 MPa in one model and 400 in the next. One legend shared by both
viewports, bottom-left above the axes gizmo, switchable from the panel.

Its maximum is **published by whoever paints the map** rather than recomputed
beside the legend: the 3D heat map samples eight points per member through the
section-stress evaluation, and deriving that twice is two answers to one
question.

The legend also **cannot outlive the picture it describes**. The publisher is
the drawing code, which stops running when the result changes — so switching
from a map to a bending diagram left the map's maximum on screen over an
unrelated picture. The value now carries the name of what drew it and is
unusable once that changes, which is what makes a stale scale impossible rather
than merely discouraged.

## 4. The pointer leaves the ribbon

Select and Pan were the exception that broke the ribbon's only rule. Every
other command opens a panel, so a lit command means "this is what the panel is
showing" — but a pointer mode shows nothing. After a solve you could have a lit
diagram, a lit Advanced and a lit drawing tool at once, three commands claiming
to be the current activity while the panel showed one.

The pointer mode is a single toggle on the model, above zoom-to-fit, in both
viewports. What stays in the ribbon is what selecting needs a panel for — which
kinds of thing a drag picks up — as **Selection**.

`isActive` is one condition now instead of five: the panel decides. Exactly one
command is lit at any moment.

Its tip states the mode you are IN, keeps the note about where the kinds are
chosen as a footnote to Select, and puts what a click would do last, behind a
rule. It is a hover tip: `:focus-within` kept it open after a click, so
`:focus-visible` replaces it, and the fade is gone because a tip consulted
mid-gesture is useless the moment it is late.

## 5. 3D → 2D: project the structure, or cut one frame out of it

**This is the largest undescribed piece of the original PR — about 1 700 lines
and a new analysis path.**

The dialog that asked what to do with the third dimension was lost in the
interface rewrite; the ribbon took its place and asked nothing, silently
switching to a 2D view of a model with a dimension it could not represent.

Reviving it as it was would have brought back its one real limitation: it only
ever projected. Flattening answers "what does this look like from the side",
and for a warehouse the answer is every frame in the building on top of one
frame with the purlins collapsed to points — 633 members becoming 140, a
picture that solves and means nothing. What an engineer wants is the frame on
grid line 12.

So the dialog asks two questions in order — **which plane**, then **what to do
with it** — and the second offers a cut. The cuts are read off the model: the
offsets where its nodes actually are, each with how many members, supports and
loads it would yield. A blank number field asks the user to remember what they
built.

A cut takes what **lies in** the plane. Members that merely pierce it are left
behind and counted. Leaving a node where each crossing member pierces the plane
was considered and rejected: it produces nodes with nothing attached, which the
solver reports as a mechanism rather than as the modelling decision it is.

Then it was measured. Every cut of every shipped 3D model — 75 of them — taken
and solved: 41 stand up, 4 are mechanisms, and **30 fail for one reason, a
frame with no support in it**. That is knowable before the cut, so the offsets
that bring no support say so on the chip, and the audit is kept as a test with
the count pinned.

@diegokingston's review found the symmetric failure and fixed it in #153, now
merged here: 71 of those 75 cuts discard some load and 16 keep members while
keeping none, including the very warehouse this module's header uses as its
motivating example. A cut with no support fails loudly; a cut with no load
solves, reports zero everywhere, and paints a uniformly green utilisation map
in the feature section 1 adds. Both are now warned about before the cut.

The conversion moved to the store, shared with the legacy toolbar PRO still
mounts — two copies of "replace the model, remember the original, remap the
plane, clear the results" is two chances to forget the backup. Coming back up
restores the 3D original; without that the trip was one-way.

## 6. Settings that say what they do

"Show primary selector" is not a sentence anyone can act on without already
knowing the answer, and it was one of twenty-five controls with nothing but its
own label. All twenty-five carry an explanation now, in three languages,
through a small `HelpTip` component.

Not the native `title`: the OS renders it in its own font at its own width,
ignores the theme, and wraps a two-line explanation into a ribbon of text
running off a narrow panel. The delay is deliberate and is what distinguishes
it from the pointer tip — these sit in a list a reader scans, and opening
instantly turns that scan into a flicker of popups.

Also: switching both selectors off used to leave the "change results view"
heading over an empty box. What the section shows is now part of whether it is
shown.

## 7. In Basic a bar is a "barra", not an "elemento"

"Elemento" is the generic word — it covers plates and quads — and Basic has
neither. 375 strings across the three languages: member in English, barra in
Spanish and Portuguese. PRO reached this conclusion first and its Members tab
already read that way.

Four things the sweep had to be kept away from, each found by reading what it
produced:

- **Reinforced concrete** — in design, code checks, detailing and revisions
  "barra" ALREADY means a reinforcing bar. One string came back reading "el
  diseño de las barras … el encaje de barras".
- **Shells** — `report.shellElements` would have become "barras shell", and the
  mesh-quality tooltips describe a surface. A bar cannot be non-planar.
- **Placeholders** — `{element}` is code. English turned it into `{member}` and
  the interpolation would have silently stopped substituting; the
  placeholder-parity test caught it.
- **Agreement** — the noun changes gender in Spanish and Portuguese, so the
  words leaning on it change with it.

`float.element` and `data.elements` are shared with Education, so that mode
reads "Member" too. Correct — an exercise is drawn out of bars — but it is a
change outside the mode this PR set out to touch, and it should be seen rather
than discovered.

## Fixed along the way

- **Selecting a member in Basic 3D highlighted nothing.** Third occurrence of
  one defect: `elementGroups` is a partial registry and in wireframe a plain
  member has no group, so `syncSelection` coloured nothing. The iterator that
  drives off the batched mesh existed already; it moves to `scene-sync` so both
  syncs share it. The rule is now enforced on the code — a loop over
  `elementGroups` may not push a colour into the batched mesh.
- `stationForces3D` read `txStart` while the solver emits `mxStart`, so every
  torsional and warping stress was zero while the panel displayed the moment.
- The stress map's gradient was missing a sign, painting the compressed flange
  red.

## Languages

41 ribbon keys and seven catalogue strings were missing from Portuguese, so the
whole ribbon fell back to English. The Basic coverage guard could not see them:
it scans for literal `t('...')` calls, and the ribbon holds its keys in a data
structure. It now also checks whole namespaces for parity — `ribbon.`, `float.`,
`selection.`, `config.tip.`, `switch2d.` — which is what catches a key no source
line mentions.

## Testing

Unit suite and the `@smoke` e2e job green, typecheck clean against the
baseline. New: 46 on box select including the permutation audit, 23 on the
plane slice, 9 auditing every cut of every shipped 3D model through the solver,
22 on the quantity/representation pair, and 42 e2e across selection, the
switch-to-2D dialog and the settings tips.

One note on the e2e job: it runs `@smoke` only, and a spec without that tag had
been failing since the pointer moved out of the ribbon without anything
reporting it. Found by running the full suite locally and fixed here.


diegokingston (commented):
Read it against `465be69`. The box-select work is clean and the stress command answers a real gap. Two things before this lands, and the first is why I want the second read carefully.

## 1. About a third of this PR is undescribed

The description covers four things — stress command, box select in 3D, colour scale, pointer mode — plus Portuguese. It does not mention:

```
web/src/components/SwitchTo2DDialog.svelte        482
web/src/lib/geometry/__tests__/plane-slice.test.ts 271
web/src/lib/store/switch-2d.ts                     257
web/src/lib/geometry/plane-slice.ts                251
web/e2e/basic-switch-2d.spec.ts                    180
web/src/lib/engine/__tests__/slice-solve-audit.test.ts 175
web/src/components/HelpTip.svelte                  108
```

~1 700 lines, none of it on `main`, and it is a **new structural-analysis path**: cut a 3D model with a plane and solve the result as a 2D frame. That is the part of this PR that most needs a reviewer's attention and the part a reviewer is least likely to know is there. Whatever else happens, it deserves its own section — or its own PR.

## 2. A slice discards load in silence

`plane-slice.ts` keeps loads whose node or element survived the cut:

```ts
const keptLoads = [...loads].filter((l) => {
  const d = l.data as { nodeId?: number; elementId?: number };
  if (d.nodeId !== undefined) return inPlane.has(d.nodeId);
  if (d.elementId !== undefined) return keptElementIds.has(d.elementId);
  return false;
});
```

That is the right rule — a load on a member the cut did not take has nothing to act on. What is missing is that **nobody counts them**. `SliceStats` reports `crossingElements` and `elsewhereElements`; `SwitchTo2DDialog` shows "N left behind" and flags a cut with no supports. Loads appear in neither.

I measured it across every cut the dialog offers on all eight shipped 3D models:

```
  cuts examined: 75      cuts that discard load: 71
  16 cuts keep members and keep ZERO load, among them:

  3d-nave-industrial   xy = 0     8 members     0 / 242 loads kept
  3d-grid-slab         xz = 0     3 members     0 / 4
  3d-space-truss       yz = 4     1 member      0 / 8
  3d-tower             xy = 3     4 members     0 / 6
```

The nave is the case your own module header uses to motivate slicing. Cut at `xy = 0` it yields eight members and not one of its 242 loads — and if that cut has supports it will **solve**, report ~zero stress everywhere, and paint a uniformly green utilisation map in the very feature this PR adds two sections above.

The asymmetry is what makes this worth fixing rather than documenting: the dialog already warns about the cut with no supports, which is the failure that is **loud** — the solver refuses and the user knows. The cut with no load is the failure that is **quiet**, and it points the reassuring way.

The fix is symmetric with what is already there: carry a `droppedLoads` count in `SliceStats`, print it beside "N left behind", and flag a zero-load cut the way `supports === 0` is flagged. Same discipline, same place, one more number.

### A latent one in the same filter

That final `return false` takes any load keyed by neither node nor element. `surface3d` carries `quadId`, so a slab or roof pressure is dropped by construction, not by the cut. No shipped fixture exercises it — I checked, they hold only `distributed3d` (230) and `nodal3d` (230), all reachable — so nothing shows it today. But a surface load is exactly what someone reaching for "switch to 2D" on a slab model would have, and the count above would at least make it visible.

## 3. Retarget: the base is already merged

This targets `audit/basic-advanced-features`, which is now fully contained in `main` (#132 merged). It should point at `main`. Against `main` it conflicts in four files — `Ribbon.svelte`, `i18n/locales/en.ts`, `i18n/locales/es.ts`, `store/ui.svelte.ts` — and everything else auto-merges, so it is a contained rebase rather than a redo.

## What I checked and found good

**Box select.** `WorldPoint` with an optional `z` rather than parallel 2D/3D types is the right call, and `Math.hypot(dx, dy, dz)` for the station along a member is the fix the description claims — a projected length would indeed have put a load at the wrong place on any member running towards the camera. The `kinds: ReadonlySet | Array` union looked like an invitation to call `.has()` on an array, but line 200 normalises through `new Set(input.kinds)` before any use, so it cannot bite.

**`planeOffsets`.** Anchoring each group on the first node's offset rather than a running centroid means the tolerance cannot drift across a chain of nearly-coincident nodes. Worth keeping as-is.

**The no-supports flag on the cut chips.** Predicting the failure before the cut instead of surfacing the solver's error after it is the right instinct — which is exactly why the load case deserves the same treatment.

Happy to push the `droppedLoads` count if that is easier than doing it yourself.

diegokingston:
Pushed the `droppedLoads` count as #153, stacked on this branch the same way #152 was on #132 — merge it in, cherry-pick it, or close it and keep the tests.

It does what the review suggested and nothing beyond it: `PlaneOffset.loads` so the dialog can warn before the cut, `SliceStats.loads`/`.droppedLoads` for after, the chip flag and warning line beside the ones you already wrote for supports, and the four strings in en/es/pt.

Two notes from writing it:

- I counted `droppedLoads` off what the SLICE kept rather than off `built.model`, because `buildSimplified2DModel` can drop a load on its own account and one number for two different decisions is how a count stops meaning anything.
- Two of your `toEqual` assertions covered the whole `PlaneOffset`, so adding a field broke them. I updated the expected objects rather than switching to `toMatchObject` — they were pinning the full shape on purpose and that is worth keeping.

The other two points from the review are untouched and yours: the ~1 700 lines of `plane-slice` / `switch-2d` / `SwitchTo2DDialog` that the description does not mention, and the retarget onto `main` now that `audit/basic-advanced-features` is merged (four conflicting files against main, everything else auto-merges).

Batuis:
Thanks — all three points were right, and the second one was a real defect.

**1. The undescribed third.** Fixed: the description is rewritten and the slice
work has its own section, flagged as the largest undescribed piece and as a new
analysis path rather than a UI change. Settings and the element→bar rename got
sections too; they were equally absent. I should have rewritten the description
when the branch grew past what it opened with instead of leaving a reviewer to
find 1 700 lines by reading the file list.

**2. The silent load discard.** Merged #153 as it stands. The measurement is
what makes the case: the warehouse this module's own header uses to motivate
slicing yields eight members and none of its 242 loads, and that cut solves.
Green everywhere, in the feature two sections above. I had built the
no-supports flag against the loud failure and then not asked what the quiet one
looked like — the asymmetry you name is exactly the gap.

Two things in your implementation I'd have got wrong: counting `droppedLoads`
off what the slice kept rather than off `built.model` (I would have folded the
builder's own drops in and produced one number for two decisions), and updating
the two full-shape `toEqual` assertions instead of loosening them to
`toMatchObject`. They were pinning the whole shape deliberately.

On `surface3d`: agreed it cannot survive a cut — a quad is not something a
plane frame carries — so dropping it is right and counting it is the fix. Worth
noting it now lands in `droppedLoads` alongside loads dropped by the cut
itself; those are different reasons under one number. Not worth splitting until
a fixture exercises it, but if slab models start reaching this path it will
want its own line rather than a bigger count.

**3. Retarget.** Done before your comment landed — base is `main`, the four
conflicts resolved. `Ribbon.svelte` and `ui.svelte.ts` were the same conflict
from both ends: main made `EDIT_TOOLS` a single export from the store while
this branch was open, so the local copies are gone and `VIEW_TOOLS` went with
the commands it described. In `en.ts`/`es.ts` main split a stress key in two
and added three; I took main's structure whole and re-applied this branch's
rename to the one string it touches there. Full suite, typecheck and the
switch-2d e2e green on the merged result.

Also re-audited the rename against the 105 commits main brought: no new
Basic-facing strings need it. The five that matched the pattern were
`solución elemental`, the DXF layer names `BARRAS/ELEMENTOS`, and the three
mesh-quality tooltips, all correctly left alone.

diegokingston (commented):
Second pass, on `fa94b0e6`. All three points from the first review are addressed, and I spent this round on the part a re-review is actually for: whether the main merge lost anything, and whether my own fix went far enough. It did not, on the last count — details below.

## The merge, checked on both sides

Four files conflicted. I checked each for content present in a parent and absent from the result:

- **`en.ts` / `es.ts`** — **zero keys from `main` lost.** Four keys sit on your side and not in the result (`footing.ui.mat{Anchorage,Geometry,LayerOrder}Pending`, `stress.noGeomMsg1`); all four exist at the merge base and were deleted by `main`, so the merge correctly took the deletion rather than resurrecting them. `stress.noGeomMsg1` in particular was split into `noGeomMsg1a`/`1b` to interpolate the section name, and `SectionStressPanel.svelte:1323` uses the pair — so the drop is right and the code agrees with it.
- **`Ribbon.svelte`** — `main`'s single-source `EDIT_TOOLS` survived: imported at line 7, used at 301 and 384, and the comment at 326 records why. Your stress command and switch-2d imports are both there.
- **`ui.svelte.ts`** — both sides present: `main`'s `export const EDIT_TOOLS` at line 17, and your `selectKinds` / `multiKindSelect` / `showColourScale` work.

Nothing lost either way as far as I can check it.

## Description

Section 5 covers it now, and putting "this is the largest undescribed piece — about 1 700 lines and a new analysis path" at the top of it is the right call. The 75-cut audit with 41/4/30 pinned as a test is a better answer than I asked for.

## What I got wrong in #153, and would like to finish

My fix warns before the cut and stops there. `SwitchTo2DDialog` shows the load count, flags a zero-load cut on the chip and warns under the outcome line — all good. But the **persistent banner** does not:

```svelte
{st.mergedNodes > 0 ? … merged …}
{st.removedElements > 0 ? … removed …}
{st.duplicateElements > 0 ? … duplicates …}
{(st.droppedCrossing ?? 0) + (st.droppedElsewhere ?? 0) > 0 ? … }
```

`switch-2d.ts:214` sets `droppedCrossing` and `droppedElsewhere` on `simplified2DStats`; it does not set `droppedLoads`, and `ui.svelte.ts:367` has no field for it. So the number exists on `SliceStats` — my change put it there — and never reaches the one place that stays on screen.

That is the wrong half to have finished. The dialog is a moment before a decision; the banner is the standing context while someone reads the results of a frame that may be carrying none of the building's load. If the count belongs anywhere, it belongs there.

It is three lines and a string: `droppedLoads` onto the stats, the field on the type, a clause in the banner. Say the word and I will push it onto this branch as a follow-up to #153, or you can take it.

## Verification

- Your CI: `test`, `suite (1)`, `suite (2)`, `e2e` and `lint` all green on `fa94b0e6` (`web` still running when I looked).
- Locally on the merged head: **6 455 passed, 1 failed**, and the one is not yours — `scene-completeness.test.ts` times out at 120 s in my sandbox because the worker pool cannot fetch, falls back to sequential, and the WASM solve alone takes 52 s. Your CI passes it. Three further files fail to *collect* here for the same class of reason (`fake-indexeddb` absent from my linked `node_modules`, Vite's `fs.allow` denying `katex.min.css?raw` through a symlink).
- `plane-slice` and the store suites: 465 passed, including the six from #153.

diegokingston:
Pushed `82f48ff8` straight to this branch — finishing #153, which I had stopped at the wrong half.

That change warned before the cut and left it there: `droppedLoads` went onto `SliceStats` and never reached `simplified2DStats`, so the persistent banner still listed merged nodes, removed elements, duplicates and dropped members with no mention of load.

The dialog is a moment before a decision; the banner is what stays on screen while someone reads the results. And the two are not symmetric there either — a frame short of **members** looks weaker than it is and a reader distrusts it, while a frame short of **load** looks stronger: it solves, reports zero, reads as safe. If the count belongs in one of the two places it is the standing one.

Three lines plus the guard: `droppedLoads` onto the stats in `switch-2d.ts`, the field on the type in `ui.svelte.ts`, a clause in the banner.

Two notes:

- The banner clause is its own element with `data-testid="s2d-dropped-loads"` and a `data-count`, because the sibling clauses are plain interpolations whose only handle is their text — and I did not want the e2e guard pinning a translated string. `basic-switch-2d.spec.ts` now asserts the count is present and positive after the Y = 10 cut on the warehouse.
- No new i18n. It reuses `switch2d.loadsLeftBehind` from #153, in the same three locales the neighbouring `switch2d.leftBehind` already lives in.

`geometry` + `store` + `i18n`: 605 passed. typecheck 479 against a 479 baseline. Build clean. I did not touch anything else on the branch.

diegokingston (commented):
Third pass, the focused one I said I owed: `result-view.ts`, the four stress measures, and both paths that paint them. This is the last part of the PR with engineering in it that I had not read.

## The `fy` fix in section 1 is real, and complete

Before, the 2D map skipped a member with no yield strength for **every** measure:

```ts
if (!sec || !mat || !mat.fy) continue;
```

Now `fy` gates only the ratio that divides by it. I checked the thing that would have made that unsafe — whether `computeElementStress` assumes `fy` exists — and it does not: `const ratio = mat.fy ? maxVM / mat.fy : null`, with σ, τ and Von Mises never touching it. So removing the guard is sound, not just convenient.

The two new measures are handled correctly too. `sigmaMax` and `tauMax` take `Math.max(|start|, |end|)` where Von Mises takes the plain max — right in both cases, because Von Mises is a square root and cannot be negative while a normal or shear stress can. And `globalMax` now tracks the absolute measures while the ratio keeps its fixed 0–1 scale, which is what makes 1.00 mean the limit regardless of the steel.

## The finding: the 2D map and the 3D map disagree, and 2D is the low one

`stress-heatmap.ts` walks 16 segments along each member and evaluates the section at each. The 2D map reads `computeElementStress`, which is handed `mStart`/`mEnd` and nothing between them — so `Math.max(vonMisesStart, vonMisesEnd)` **structurally cannot see a peak inside the span.**

Measured on the most ordinary beam there is — simply supported, IPE 300, L = 6 m, w = 20 kN/m — using this repo's own `computeElementStress` for one number and its own `computeDiagramValueAt` for the other:

```
end-only vonMises (what the 2D map paints):  29.0 MPa
peak moment found by sampling:               90.0 kN·m
bending stress there:                       161.6 MPa
under-report factor:                          5.6x
utilisation painted vs actual (fy = 275):   0.11  vs  0.59
```

A beam at 59 % of yield paints as 11 %. For a simply supported member the ends carry no moment at all, so the map is reading the one place along the member where bending is zero.

**This is pre-existing** — the two-point read predates this branch, and I am not asking you to own it. Two things make it worth raising now rather than later:

1. This PR extends the same read to `sigmaMax` and `tauMax`, so the flaw arrives in two more measures.
2. This PR gives both viewports a **numeric legend** whose maximum is, in your words, "published by whoever paints the map". The two painters now publish different maxima for the same model and the same measure, and a user can read one, switch viewport, and read the other. What used to be an invisible weakness is now a visible contradiction.

The direction is the unsafe one, which is the only reason I am spending words on a pre-existing issue.

I would **not hold this PR for it**. The fix means sampling the 2D map the way the 3D one does — `computeDiagramValueAt` for N, V and M at each t, then the section evaluation — which is a real change to a hot drawing path with its own correctness questions, and it does not belong in a branch this size. A follow-up issue with the numbers above attached is the right home for it. Happy to write it.

## Two trivia in `result-view.ts`

- `commandShowsQuantity` has a dead branch: `if (diagram === 'axial') return active === 'axial';` is exactly the `return diagram === active;` on the next line.
- `showQuantityAs(q, 'memberColour')` for an unsigned quantity silently draws a diagram instead. `representationsFor` never offers that combination so the UI cannot reach it, but the function is exported and the fallback is quiet.

Neither is worth a commit on its own.

## Where that leaves my review of this PR

Across three passes I have read the slice path and its tests, box-select, the store wiring, the four conflicted files of the main merge, and now the stress measures and both painting paths. What I have **not** read: `PointerModeButton`, `SelectionPanel`, `ColourScaleLegend`, `HelpTip`, the two viewport components beyond the map code, `ToolbarResults`, the 297 lines of new box-select tests, and the three new e2e specs — presentation and coverage, where ten minutes of someone clicking through Basic is worth more than me reading Svelte.

On what I did read, I have nothing outstanding. It is green, it contains all of `main`, and I would merge it once someone puts an actual approval on it — mine are comments, and a 42-file PR that reaches the solver through the slice deserves more than that as its review record.

PR #154, opened 2026-08-19
review fixes for #151: legend units and ramp, switch-2d crash and backup, slice releases

diegokingston:
# Review fixes for #151

These are the fixes from the review of #151, prepared and verified against its head — but #151 merged (`82f48ff8`) before this commit was pushed, so it lands here instead. One commit, `7de2d59b`, 46 files, +1368/−188.

## Colour scale and stress maps

- The 3D heat map published its maximum in kPa under an "MPa" label — the legend read 235000 MPa for a member at 235 MPa. Stress maxima are converted kPa→MPa at publish time only; internal normalization stays in native units. Pinned by test.
- `getSectionProps` passed `mat.fy` (MPa) into `computeSectionStress` (kPa), so 3D utilisation was ~1000× off for any material with fy set; fy is converted, and members without fy are skipped in 3D exactly as in 2D instead of being painted against an assumed 355 MPa. The no-yield warning now fires when ANY painted member lacks fy, not only when none does.
- The legend ramp matched neither painter: one shared ramp (`lib/three/colour-ramp.ts`) now feeds the legend, the 3D `heatmapColor` and the 2D canvas formula. Utilisation >1 paints magenta in both viewports and the legend shows the "> fy" cap entry. `heatmapUnit`/`colourScaleUnit` are one export now.
- The measure select unmounted itself for shell contours and no ribbon command lit while one was shown; the select is gated on `activeMapMeasure` and the Stress command lights on the same predicate.

## Switch to 2D

- Typing in the slice offset field crashed the dialog: `bind:value` on a number input coerces to `number|null` and `offsetText.trim()` threw. The field keeps a string via `value` + `oninput`.
- The module-level 3D backup survived project loads: slice model A, load project B, press dim-up and `restore3D()` wiped B. `resetSwitchBackup()` is wired into `modelStore.clear()` and `deserializeProject()`; a test proves a load can no longer be replaced by a stale backup.
- Slices silently changed connectivity for released members: the 2D solver reads only `mz` as the bending hinge, so an XZ frame's in-plane `my` releases vanished and its out-of-plane `mz` became a phantom hinge. The shared builder now remaps the release about the axis normal to the target plane into `mz` (XZ: my→mz, YZ: mx→mz) and drops the rest, for the slice and project paths alike. Per-plane tests.

## Selection and pointer

- An empty 3D marquee in supports/loads mode kept the old selection highlighted and Delete-armed while nodes/elements cleared in the same handler; non-additive gestures now clear all channels.
- The 3D click ignored the multi-kind setting `PointerModeButton` exposes there; it now mirrors the 2D branch (nearest among all armed kinds, channels cleared up front). `PointerModeButton` shows the armed build tool instead of claiming "Select", and no longer renders a blank icon for influenceLine.
- Tool shortcuts (N/E/S/L) sync the data tab the way DataTable does in the other direction, so exactly one ribbon command lights.
- box-select gains a concrete-value test pinning the true-length station; stale header and `SelectMode`-narrowing casts fixed on both viewports.

## i18n and hygiene

- pt `toolbar.planeModal.simplified` said "elementos" after the elemento→barra rename; now "barras", and the full 24-key `switch2d.*` block exists in all 13 locales.
- Tour strings no longer reference the retired zoom-to-fit glyph in any locale (also repaired two corrupted characters in hi/tr).
- ToolbarConfig no longer stacks a native `title` tooltip on top of a HelpTip; HelpTip's tooltip is programmatically associated; SelectionPanel uses `role=group` + `aria-pressed`; dead Toolbar imports and stale comments removed.

## Verification

- Full unit suite green: 6483 tests + the 35 autosave tests (initial run failed only on a `node_modules` that predated the `fake-indexeddb` devDependency; 35/35 after `npm install`).
- `typecheck`: 479 errors reported, baseline 479 — no new type errors.
- `check:gate`: no errors in guarded paths.
- Rebased onto the post-#153 branch head before landing here; overlap with #153's dropped-loads fix resolved in its favor where it subsumed these changes (the release remapping and per-plane tests here are the part #153 does not cover).


diegokingston (commented):
Read it against `7de2d59b`. Ran the affected suites: `geometry` + `three` + `switch-2d` + `viewport3d`, 451 passed.

Two of these fixes overlap work I had in flight, and **yours are better on both counts** — I am dropping mine. Details below, then the one gap I found and one that is still open.

## The fy and units fix supersedes mine

I had the same two bugs open on a branch and was converting in `sampleValue`. Converting at **publish time** instead is the better call, and the comment says why in one line: "the one place the number and its unit are decided together". Keeping the internal normalisation in native units means the painted gradient and the legend cannot drift apart, which my version left possible.

Two things you did that I had not:

- `fy: mat?.fy ? mat.fy * 1000 : null` and then **skipping** the member, rather than my `?? 355_000` fallback. Grading a concrete member against an assumed steel strength is worse than not painting it, and I had left that in place.
- Flipping `if (m?.fy) return true` → `if (!m?.fy)`. That warning fired only when **no** member had a yield strength, i.e. never in a mixed model — which is exactly when it was needed. I missed it entirely.

The one thing that survives from my branch is a different bug in the same area, in the Rust engine rather than the web: `section_stress.rs`'s CHS branch still pairs a one-sided `Q = t(R²−y²)` with a two-sided `b = 2t`, so a real tube reports `V/A` where the TS closed form gives `2V/A`. Measured on a CHS 300×8: engine 1.05× the mean, TS 2.11×, thin-wall exact 2.0. That is unrelated to this PR and I will send it separately.

## The gap: `inPlaneBendingAxis` ignores two orientation overrides

`plane-projection.ts:126` replicates the auto-orient convention to decide whether the in-plane hinge is `my` or `mz`. `computeLocalAxes3D` (`local-axes-3d.ts:66`) supports two things the replica does not:

- **`localY`** — an explicit orientation reference, which takes a completely different branch (`ez = normalize(ex × ref)`);
- **`rollAngle`** — a rotation of ey/ez about ex. `modelStore.rotateElementLocalAxes` is a real action and `KeyboardShortcuts.svelte:147` calls it.

The call site passes geometry only:

```ts
const axis = inPlaneBendingAxis(plane, b.x - a.x, b.y - a.y, b.z - a.z);
```

So for a member with an explicit local-Y or a non-zero roll, the axis it picks is the auto-orient one, not the member's. At a 90° roll `my` and `mz` swap outright — which puts the out-of-plane release into `mz` and drops the in-plane one. That is the same defect this fix exists to prevent, surviving in the rolled case.

The deeper point is the duplication itself: the replica is a second copy of a convention that lives elsewhere, kept in step by a comment ("see engine/local-axes-3d.ts") rather than by the compiler. Calling `computeLocalAxes3D` and comparing its returned `ey`/`ez` against the plane normal would be shorter, would pick up `localY` and `rollAngle` for free, and could not drift.

I have not built a model with a rolled member sliced out of a frame to confirm the end-to-end consequence — the code path is clear enough that I think it is worth closing regardless, but treat the severity as unmeasured.

## Still open: the zero-load flag does not fire on 7 shipped cuts

Adding the builder's `droppedLoads` to the slice's is right, and it closes the "element removed" and "3D-only type" cases. It does not close the one that matters, and I ran it against this head:

```
3d-torsion-beam    xy=0     advertised=1  kept=1  dropped=0    magnitude=0
3d-space-truss     xy=1.5   advertised=8  kept=8  dropped=0    magnitude=0
3d-grid-slab       xy=0     advertised=4  kept=4  dropped=0    magnitude=0
3d-nave-industrial xy=8.5   advertised=40 kept=40 dropped=202  magnitude=0
3d-nave-industrial xy=9     advertised=40 kept=40 dropped=202  magnitude=0
3d-nave-industrial xy=9.5   advertised=40 kept=40 dropped=202  magnitude=0
3d-nave-industrial xy=10    advertised=20 kept=20 dropped=222  magnitude=0
```

Seven cuts advertise between 1 and 40 loads, keep all of them, and produce a 2D model carrying **zero total magnitude**. The zero-load warning is gated on `c.loads === 0`, so it stays silent — the frame solves, every result is zero, and the utilisation map paints it uniformly green.

The mechanism is at `plane-projection.ts:338-344`: a `distributed3d` keeps only its in-plane component and the object is **pushed regardless**, so a vertical roof load cut on a horizontal XY plane becomes `qI = qJ = 0` and is counted as kept. The nodal path at `:363` has the mirror case — a sum below 1e-15 is `continue`d without incrementing `droppedLoads`.

This is **my bug, from #153** — I counted load objects rather than load, and the commit message claimed the opposite. I am not asking you to fix it; I am flagging that this PR does not, so it does not read as closed. I will send the magnitude-based count.

## Smaller things

- The `matching` tolerance going 1e-6 → 1e-3 with the "fmt rounds to three decimals" reasoning is right, and the note that it equals `SLICE_TOL` so two cuts cannot both match is the part worth having written down.
- `PROJECTION_COLLAPSE_ERROR` as a named export beats matching an English sentence at the call site.
- `resetSwitchBackup()` wired into `clear()` and `deserializeProject()` closes a genuinely nasty one — a stale backup overwriting a freshly loaded project is the kind of bug people report as "it ate my file".
- The `value` + `oninput` fix for the numeric input is right, and the comment explains the coercion clearly enough that nobody will "simplify" it back to `bind:value`.

## Verdict

The switch-2d crash, the backup reset, the fy/units work and the release remapping are all worth landing, and three of the four are things I either missed or did worse. I would fix `inPlaneBendingAxis` to call `computeLocalAxes3D` before merge, since the replica is new code and a rolled member is reachable today. The zero-load counting is mine to finish.

diegokingston (commented):
Re-reviewed at `7de2d59b` — unchanged since my last pass, so everything there still stands. What I did this round is **measure the one finding I explicitly left unmeasured**, because "the code path is clear enough" was not good enough for something I was asking you to change before merge.

It reproduces, and the consequence is worse than I guessed.

## Measured: a rolled member silently loses its in-plane hinge

`inPlaneBendingAxis` replicates the auto-orient convention and receives geometry only. `computeLocalAxes3D` — the function the solver actually uses — also honours `rollAngle`. Asked at several rolls, for a horizontal member in the XZ plane:

```
roll      canonical in-plane axis
   0°     my
  30°     my
  45°     my
  90°     mz     ← the replica still says my
 180°     my
```

At 90° the two disagree. Driving a real slice with a member rolled 90°, whose in-plane hinge is therefore stored as `mz`:

```
canonical says the in-plane axis is:  mz
slice produced releaseI:              { my: false, mz: false, t: false }
→ the hinge WAS LOST
```

The remap keeps `axis === 'my' ? r.my : r.mz`; with `axis` wrongly `'my'` it copies `r.my` (false) and discards `r.mz` (the real hinge). The 2D frame comes out with moment continuity where the structure has a pin — **stiffer than the real thing**, redistributing moment into a joint that cannot carry it. Same class of defect this fix was written to eliminate, surviving in the rolled case, and silent.

`rotateElementLocalAxes` is a real store action reachable from the keyboard shortcuts, so this is not hypothetical geometry.

Neither `rollAngle` nor `localY` appears anywhere in `plane-slice.ts` or `plane-projection.ts`, so the replica cannot see either.

## What I would change

Call `computeLocalAxes3D` and compare its returned `ey`/`ez` against the plane normal, instead of re-deriving them. Shorter than the replica, picks up `rollAngle` and `localY` for free, and cannot drift from the convention it is imitating — which is the deeper problem: right now a comment ("see engine/local-axes-3d.ts") is the only thing holding two copies of a convention in step.

I did not push this myself: it is inside the release-remapping logic you wrote, the right behaviour for `localY` needs a decision I would be guessing at, and you may prefer a different shape.

## Everything else from the first review still stands

Still worth having, unchanged: the `fy`/units work supersedes what I had (and the `if (!m?.fy)` flip is a bug I had missed entirely), `resetSwitchBackup` closes a genuinely nasty one, and the `value` + `oninput` fix is right.

Also unchanged: the seven silent zero-load cuts. That is **my** bug from #153, not yours — the fix is #158, stacked on this branch and now green.

PR #158, opened 2026-08-21
fix(slice): count load by what it carries, not by what survives the cut

diegokingston:
Finishing my own #153, which did not do what its commit message claimed. Targets `basic/stress-maps-and-selection` (#154's head), since that is where the slice code currently lives.

## What was wrong

#153 counted load **objects**. A load survives a cut whenever the thing it acts on survives — but it carries only its **in-plane component** into the 2D frame. `plane-projection` kept the in-plane part and pushed the object regardless, so a roof load pointing down global Z, cut onto a **horizontal** frame, became `qI = qJ = 0` and was counted as kept.

Measured over every cut of every shipped 3D model:

```
3d-torsion-beam     xy=0     advertised  1 load   → frame carries 0
3d-space-truss      xy=1.5   advertised  8 loads  → 0
3d-grid-slab        xy=0     advertised  4 loads  → 0
3d-nave-industrial  xy=8.5   advertised 40 loads  → 0
3d-nave-industrial  xy=9     advertised 40 loads  → 0
3d-nave-industrial  xy=9.5   advertised 40 loads  → 0
3d-nave-industrial  xy=10    advertised 20 loads  → 0
```

The zero-load warning is gated on that count, so it stayed silent on exactly the seven cuts that needed it. The frame solves, every result is zero, and the utilisation map paints it uniformly green — the quiet failure #153 exists to prevent, still there and now carrying the counter's approval.

#154 added the builder's own `droppedLoads` to the slice's, which closes the "element removed" and "3D-only kind" cases. It does not close this one, and I verified that against its head before starting.

## What this does

- **`inPlaneLoadMagnitude`** — one definition of what a load carries in a plane, used by the preview *and* by the builder's drop decision, so the two cannot answer differently.
- A distributed load projecting to zero is **dropped and counted**, not pushed as an empty object. Nodal sums that vanish, and sums on a node that did not survive, are counted too — by **how many source loads fed them**, because three loads merging into one that then drops is three lost, not one.
- **`stats.carriedLoads`** counts source loads rather than emitted ones, since the builder merges nodal loads at a shared node. Counting the output would report losses that never happened — which is how the conservation check caught this while I was writing it.

## A claim I had to withdraw

#153 left a comment saying the number shown before the cut equals the number the cut honours. **It does not**, and four cuts of the library disprove it: the preview runs before the projection, so it cannot know which members will collapse or duplicate away, and a load on one of those goes with it. Making it exact would mean running a full build per candidate offset — 25 of them on the nave alone.

So the test asserts the equivalence the warning actually depends on, in both directions: **zero advertised if and only if zero carried**. That is deliverable and it is what protects the user; an equality I cannot deliver would have been another comment that reads true and is not.

## Verification

- Library sweep added beside the existing cut audit, reusing its fixtures and mock rather than duplicating them. Over 75 cuts: **silent zero-load 7 → 0**, zero-disagreement 0, and `loads + droppedLoads` equal to the model's own total on every one.
- Reverting the zero-magnitude drop turns the sweep red — checked.
- `geometry` + slice audit + `store`: 562 passed, 0 failed. (Two files fail to *collect* in my worktree: `fake-indexeddb` postdates the `node_modules` I have linked.)
- `npm run typecheck`: no new errors (479, baseline 479).

## Note

`planeOffsets`' `loads` parameter now needs `type` as well as `data` — what a load carries depends on its kind, so the count cannot come from the payload alone. Every caller already passes whole load objects.

### DIFF
--- a/web/src/lib/engine/__tests__/slice-solve-audit.test.ts
+++ b/web/src/lib/engine/__tests__/slice-solve-audit.test.ts
@@ -177,3 +177,73 @@ describe('every cut of every 3D model', { timeout: 120_000 }, () => {
     expect(solved / total).toBeGreaterThan(0.3);
   });
 });
+
+/**
+ * Load accounting, over the same library.
+ *
+ * A cut that keeps its supports fails LOUDLY when it is wrong — the solver
+ * refuses. A cut that keeps no LOAD fails quietly: it solves, every result is
+ * zero, and the utilisation map paints it uniformly safe. The dialog warns
+ * about that, and the warning is only as good as the count it is gated on.
+ *
+ * That count used to be a count of load OBJECTS. A load survives a cut
+ * whenever the thing it acts on survives, but it carries only its in-plane
+ * component into the frame — so a roof load pointing down global Z survived a
+ * horizontal cut and acted on nothing. Seven cuts of this library advertised
+ * between one and forty loads and produced a frame carrying zero, with the
+ * warning silent on every one.
+ */
+describe('load accounting across the library', { timeout: 120_000 }, () => {
+  /** Everything the produced 2D model actually carries, summed. */
+  const magnitude = (loads: Array<{ data: Record<string, unknown> }>): number => {
+    let m = 0;
+    for (const l of loads) {
+      for (const k of ['qI', 'qJ', 'fx', 'fz', 'p', 'my']) {
+        m += Math.abs(Number((l.data as Record<string, unknown>)[k] ?? 0));
+      }
+    }
+    return m;
+  };
+
+  it('no cut advertises load and delivers a frame carrying none', () => {
+    let checked = 0;
+    for (const name of MODELS) {
+      const model = load(name);
+      if (!model.loads.length) continue;
+      for (const plane of ['xy', 'xz', 'yz'] as DrawPlane[]) {
+        for (const cut of planeOffsets(
+          plane, model.nodes.values(), model.elements.values(),
+          model.supports.values(), model.loads,
+        )) {
+          if (cut.elements === 0) continue;
+          const r = sliceModelAtPlane(
+            plane, cut.value, model.nodes.values(), model.elements.values(),
+            model.supports.values(), model.loads, model.materials, model.sections,
+          );
+          if (!r.ok) continue;
+          checked++;
+          const where = `${name} ${plane}=${cut.value}`;
+
+          // The one that matters: if the dialog says there is load, there is.
+          if (cut.loads > 0) {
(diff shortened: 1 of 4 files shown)


diegokingston (commented):
Reviewed my own PR and found a real defect in it. Fixed in `6c67938a`, pushed here.

## The claim that was not true

The description says `inPlaneLoadMagnitude` is "one definition of what a load carries in a plane, used by the preview **and** by the builder's drop decision, so the two cannot answer differently."

That was true of the `distributed` branch and of nothing else. `pointOnElement` and `thermal` got the element-survival check and no magnitude check at all:

```
case                     advertised   slice.loads   dropped   AGREE?
pointOnElement p=0              0             1         0   *** NO ***
thermal all zero                0             1         0   *** NO ***
pointOnElement p=5              1             1         0   yes
thermal dt=20                   1             1         0   yes
distributed q=0                 0             0         1   yes
nodal all zero                  0             0         1   yes
```

A point load of 0 kN was advertised as nothing by the preview and counted as **carried** by the projection, so `slice.loads` reported "1 load" about a frame carrying none. That is precisely the lie this PR exists to stop, in a rarer type — and the fix that was supposed to make it structurally impossible was applied to one branch out of three.

## Why the library sweep could not catch it

This is the part worth keeping. The sweep passes over all 75 cuts of all eight shipped models, and it proved nothing here, because **no shipped fixture holds a zero-magnitude point or thermal load**. A sweep over real models only ever proves things about the loads real models have.

So the new tests enumerate the KINDS rather than the models: each of the six load types at zero, and four of them carrying something. The second half matters as much as the first — without it the rule could be satisfied by dropping everything, which silences the warning just as effectively as counting everything.

Removing the guard from any single branch turns that branch's own case red. Checked on `pointOnElement`.

## Now

- The check is one closure applied at every element-attached branch, so it cannot be applied unevenly again.
- Nodal loads stay out of it deliberately: several can land on one merged node, and two that each carry something can still sum to nothing, so that decision belongs on the sum and is made there.
- `geometry` + slice audit + `store`: 572 passed. typecheck 479 against a 479 baseline.

## What still stands from the original description

Everything else in it holds, and I re-ran it: the seven silent zero-load cuts go to zero, conservation holds on all 75, and the withdrawn claim about advertised-equals-kept is still withdrawn — the preview runs before the projection and cannot know which members collapse or duplicate away.

One thing I have NOT verified and should say plainly: I have not checked whether a nodal load that carries something individually, but whose merged sum cancels to zero, is counted the way I would want. Two equal and opposite loads at one node is a strange model, and I did not build one.

PR #164, opened 2026-08-22
M2 — PRO steel: section and material selectors, generators, the metallic workflow, and joint design

Batuis:
> ## Ready for review — read this box first
>
> **This PR depends on #156. Merge #156 into `main` first, then this one.**
> The base of this PR is `feat/pro-steel-m1`, and this branch **contains M1 by merge**. It is not
> independent of #156. Do **not** retarget it at `main` to merge it earlier: that would carry M1's
> entire diff in under this PR's review.
>
> **Scope of this PR** — the PRO steel workflow on top of M1: section and material selectors behind
> a single modal, the generators, joint design, and I-06 / I-07 / I-08 (joints persist, stay with
> their model, and now travel on a share link).
>
> **No new functionality was added while preparing this for review.** The only changes since the
> feature work are documentation and two inherited test fixes.
>
> **One open product decision, which does not block: A-01** — see
> [A-01](#a-01--the-one-open-decision-and-it-is-not-a-blocker) and
> `docs/handoffs/a-01-decision.md`.
>
> Jump to: **[Delivered scope](#delivered-scope)** · **[Known limitations](#known-limitations)** ·
> **[Non-blocking findings, for M3](#non-blocking-findings-for-m3)** ·
> **[Nothing classified as blocking](#nothing-classified-as-blocking)**

## Stacked on M1 — **do not merge before M1**

This branch starts from `feat/pro-steel-m1@f936f29c` and its base is **`feat/pro-steel-m1`**, not
`main`. Intended order: **M1 (#156) → M2 (this) → M3**. Both PRs stay in draft and stay separate.
Neither is closed, neither is rebased, neither has its base changed.

---

## What M2 delivers

Metallic (steel) work in PRO, from the section a user picks to the joint they design.

### 1 · The PRO section selector

Two divisions and no third one. A section is either **chosen from a catalogue** or **built from a
named template** — there is no amorphous option where a user types an area and an inertia with no
geometry, because such a section cannot be drawn, composed, classified or checked against a
clause. Composition, gap and rotation live in one modal, so composing a back-to-back angle for a
generator role and composing one in the sections tab are the same act on the same object. The
cold-formed C/Z work sits inside that selector rather than as a workflow stage.

### 2 · The PRO material selector

`gradeId`, `standard`, `region` and `fu` persist. Aluminium, timber, concrete and steel classify
correctly, and the panel distinguishes an identified grade from a typical value outside the
metallic catalogue and from an absent grade. Each field on the data sheet carries its provenance.

### 3 · Generators

Docked preview, profile and material rows that reuse the selectors above, and truss geometry that
matches its name: **Pratt** diagonals descend toward midspan, **Howe** is the inverse, **Warren**
has no posts. The subdivision option is a checkbox and is not called "Baltimore" in the interface.

### 4 · The metallic workflow, in five stages

The regulation gate depends on `steelCodeDeclared`, not on whether CIRSOC 301 is mature enough to
certify. A stage that cannot be opened says which of its prerequisites is missing.

### 5 · Joint design

Not a minimum envelope dressed as a designed joint. Per-joint demands are walked across every
combination and never mixed between them, with the governing demand reported per component
together with the combination, element and end it came from.

Implemented with the clause attached to every number: §J.3.2/Table J.3.3 (hole sizes), §J.3.3
(`s ≥ 3d`), §J.3.4/Table J.3.4 (minimum edge distance), §J.3.5 (maximum spacing), §J.3.6/Table
J.3.2 (bolt strength), §J.3.7 (combined tension and shear), §J.3.10 (bearing), §J.2.1 (end-loaded
β), §J.2.2 (throat and leg limits), §J.2.4/Table J.2.5 (weld metal), §E.6.1 (the five groups),
§E.6.3.1 and §E.6.3.2 (batten stations and chord unbraced length), §E.6.19 (the stiffness
condition).

**What is deliberately not produced:** batten plate thickness, width and depth. §E.6 names no
batten dimension anywhere — the only property it gives is `Ip`, and only inside `np·Ip/h ≥ 10·I1/a`.
Those come back as `GEOMETRY_UNAVAILABLE` with the condition they would have to satisfy, never as
an invented dimension. Fillet welds top out at `notVerifiable` because Table J.2.5 refers the base
metal to Chapter J.4, which is documented as a bounded future extension (tension, shear and block
shear of the connected elements) rather than implemented here.

`designed` is never presented as approval.

### 6 · 3-D visualisation

Selection is one channel in both directions: selecting a node opens its joint in the panel, and
selecting a joint in the panel selects the node. A joint with no producible geometry draws
nothing rather than drawing a placeholder.

---

## Two classes of defect this branch found

**A plausible value standing in for absent data.** Three times, under three different names:

- The shed fixture's second material had no `fy`, so all 633 elements classified as `unknown` and
  the joint inventory reported 0 of 633. Fixed in the data, not in the predicate.
- `getJointForces` read `NI`/`NJ`; the real fields are `nStart`/`nEnd`. Every lookup returned
  `undefined` and fell through to `?? 0`, so the connections panel showed a table of zeros for
  every joint in every model — and a zero force reads as an unloaded member, not a missing one.
- The batten gap was read as `Number(value) || 10`. Zero is falsy, so a deliberate 0 became 10 and
  §E.6.1's Group I — chords in continuous contact, carrying no battens — became unreachable while
  the section went on claiming battens for it.

The third is now generalised: one parser keeps a real zero, an empty field and an invalid entry
apart, every field declares what zero means for it, and no `Number(...) || default` remains in
`src/components/pro/`.

**A correct state a user could not leave.** The batten form lived inside the "layout available"
branch, so typing a gap of 0 unmounted the very control that had just been used.

---

## Gates

| Gate | Result |
|---|---|
| `npm run test:unit` | **7747 pass**, 12 skipped, 1 todo |
| `npm run test:build` | **14 pass** |
| `npm run typecheck` | **473**, no new errors against the baseline |
| `npm run build` | exit 0 |
| i18n es/en/pt | **784 keys, at parity** |

**The renderer crash, found and mitigated.** About one test in ten that loaded the shed and
solved it failed with `page.evaluate: Test timeout of 60000ms exceeded`. It was neither a hang nor
the solver: **the page crashes**, Chromium aborts with `signal 4, ILL_ILLTRP`, and `page.evaluate`
on a crashed page does not reject — it waits. That one fact made three readings of the symptom
wrong at once: the page's own `setTimeout` not firing, a 60 ms solve never returning, and the
sequential fallback never running.

Localised by measurement, one variable at a time, 40 sessions per arm — pool disabled 0/40, one
worker 0/40, small model 0/40, loaded but never solved 0/40, workers created serially 7.5 %, one
compiled module per worker 12.5 %, `--no-wasm-tier-up` 15 %, full Chromium channel 7.5 %, Chromium
151 22.5 %, trace and video off unchanged, peak RSS 859 MB against 922. So not the GL backend, not
memory, not module sharing, not the WASM tier-up, not the browser build. **In real Google Chrome
the rate is 17.5 %**, so a user solving a large model loses the tab: a product defect, not a test
artefact.

The decisive measurement: 60 sessions solving three times each — 180 solves — gave 7 crashes and
**all seven were on the first solve**; none of the ~113 with already-executed workers crashed. The
trigger is one worker's first execution coinciding with another's.

**Fixed in `solver-pool.ts`**: while any reachable worker has never executed, the first job on
each runs one at a time on a distinct worker; once they have all executed once, dispatch is
simultaneous again.

| | first solve | later solves | crashes |
|---|---|---|---|
| before | 59–61 ms | 44–47 ms | 2 in 12 sessions (16.7 %), both on solve 1 |
| after | 144–150 ms | 44–46 ms | **0 in 160 solves** |

About 88 ms once per session; the session median is unchanged and the steady state improves.
It removes the condition measured to trigger the crash — it does not repair the underlying fault,
which is in the WASM execution path or in V8, outside what this branch may touch. A regression
test asserts the invariant (no cold job is handed over while another is in flight, each warms a
different worker, and later solves overlap again) rather than a duration.

Two supporting fixes: `withCrashGuard` makes a crash fail immediately and name itself instead of
costing 60 s and pointing at the wrong line, and `Viewport3D` now releases its WebGL context on
teardown as `RebarViewport3D` always did. An M1 spec that M2 had broken was also corrected — it
asserted on a generator row that no longer carries the arrangement.

No timeout was raised, no retry added, no test marked flaky, no WebGL disabled. Full detail in §9
of `docs/handoffs/m2-qa-handoff.md`.

---

## Scope boundaries held

No changes to the solver, Rust, Cargo or the WASM sources. `StageSection` untouched. No rebase.
No snapshots updated and no timeouts inflated.

---

# The three open findings, decided and closed

§6 of `docs/handoffs/m1-m2-audit.md` left three product decisions open. All three are now
implemented, one commit each, preceded by a defect the first of them uncovered. Full account,
including the CI audit and what is deliberately *not* closed:
`docs/handoffs/m1-m2-ci-audit-and-three-decisions.md`.

## 0 · A section built through the modal had no geometry — `806e1289`

Measured before deleting anything, because removing the tab's inline form without this would have
replaced a working path with a broken one.

`computeSectionProperties` returns `tw`, `tf`, `t` and `tl`. The `built` branch of
`toSectionFields` copied `a`, `iy`, `iz`, `j`, `b`, `h` and `shape` and stopped — `SectionFields`
had no thickness fields at all. For a catalogue pick that is right: the resolver finds the entry
by name and reads the published dimensions off the catalogue. A built section has no entry, so it
switches on `shape` and calls `need('b','h','tw','tf')` against the section itself.

Same profile, both paths:

| Path | Canonical state |
|---|---|
| Modal (`toSectionFields`) | `properties-only`, `missing: ['tw','tf']` |
| Inline form | `geometry-backed` |

Undrawn, unextruded and outside every clause helper that dispatches on shape. `hollow-rect`, the
first template in the list, already fails it. Nothing caught it because the properties are
identical either way — area, inertia, mass and every solver result agreed, and only the geometry
differed.

## 1 · One catalogue for a profile, not two — `4a458b39`

`ProSectionsTab` goes from 724 lines to 205. The family strip, the search box, the
`<tr onclick>` table and the inline build form are gone; the modal trigger is the only way in.

The inline path was the poorer by four measures a user could hit: it called
`modelStore.addSection` directly and so produced a section with no arrangement, gap or rotation;
its rows had no tab stop, no Enter and no role; it offered no standards body, design code or depth
filter; and the data sheet existed only on the modal.

Verified before deleting: every consumer of the shared class names inventoried (the E2E that
clicks `.profile-row` drives `/app/basic`, which is `SectionChanger`); generators and the PRO tab
reach the same `ProSectionModal`; all fifteen families still produce a section with an area *and*
with `profileFamily`, walked from `FAMILY_LIST` rather than from a number written down; and no
solver, engine or calculation model is touched.

`ColdFormedPanel` is deliberately left alone. It creates sections too, but the parametric C/Z
catalogue is not one of the fifteen and has no modal equivalent — removing it would lose a
capability rather than a duplicate.

> Beyond the written proposal: it said the builder was not to be touched. It was, because the
> instruction was that the modal be the only source of *creation* and `BuiltSectionPanel` reads
> exactly the same `SECTION_SHAPES` lists the inline form did. Flagged so reverting it is a
> decision rather than a discovery.

## 2 · The auxiliary calculation stops giving a verdict — `8e538631`

The tick was in two places, and finding the second is what made the first worth fixing:

- **the result card** — `conn-status-icon` mapped `ok` to `✓` on `--st-ok`, and the weld card
  added two more on its size range and its `L ≥ 4a` condition;
- **the section header** — `StageSection` paints `done` as a ✓ in `--st-ok`, and both sections
  reached `done` the moment a result object existed. Not when the result was good: when it
  *existed*. A bolt group over its capacity turned its own header green, and so did pressing
  Verify on the defaults, where Vu and Tu are 0 and the utilisation is 0 % because nothing was
  asked of the joint.

Converted rather than deleted. The outcome is stated in words of its own —
`within / near the limit / over the limit`, never the canonical `adequate`, whose Spanish is
«cumple» and one word from the collision — there is no success tone in the mapping to reach for,
each card says what it is before it shows a number and names where the state that counts comes
from, and the sections are `optional`, never `done`. §1, joint *detection*, keeps its `done`
deliberately: it means the detector ran, and says nothing about adequacy.

`ProConnectionsTab`'s exemption from the `steel-never-verified` sweep is gone. `ProVerificationTab`
keeps its own, and only it, for the reason it always had: that tab genuinely shares a component
with reinforced concrete.

> Beyond the written proposal: it suggested reusing the metallic status badge. The four metallic
> states describe a member's design status, not a utilisation ratio, and reusing them is the
> mixing of vocabularies the decision asks to avoid.

## 3 · The section figure joins the token system — `4b0afd2b`

It was the only metallic component with hardcoded colour. Each replacement was measured, and one
came out against the obvious answer:

| | before | after | measured |
|---|---|---|---|
| Well and void fill | `#071322` ×2 | `var(--st-bg)` | 1.14 → 1.11 against the row; `#071322` vs `--st-bg` is **1.02** |
| Frame | `#24486e` | `var(--st-hair-strong)` | 1.74 → **2.03** on the row, 1.98 → 2.07 on the well |
| Empty-state em dash | `#566` | `var(--st-text-2)` | **3.02 → 7.00**, closing a pre-existing AA failure |

`--st-hair` is the obvious reading and the wrong one: 1.48, *below* the literal's 1.74, so the
frame would come out fainter than before. Inside the modal this figure also sits within a
`--st-hair` preview well, and a nested frame in its parent's own token is one nobody sees.

No new token: `tokens.css` is H1's. **No snapshot was updated and none needed to be** — the only
two committed baselines are `overlay-legend` and `batch-dialog`, both reinforced-concrete surfaces
in `rc-design-visual.spec.ts`, and neither renders this component.

## Gates

7977 unit tests and 14 build tests pass, typecheck is at its 473 baseline with no new errors, the
production build exits 0, i18n parity across es/en/pt holds with the seven new keys, and the full E2E suite is
**800 passed / 5 failed / 1 flaky** across 67 specs in 43.2 min. Every one of the five has a twin
on a clean `origin/main` checkout whose failure text is identical — see §19 of
`docs/handoffs/m1-m2-audit.md`. Every run used a dedicated port.

## Scope, again

No solver, Rust, Cargo or WASM sources. `StageSection` itself untouched — only the state
`ProConnectionsTab` hands it. No rebase, no force-push, no snapshot updated, no timeout inflated.

---

# The five E2E failures, classified against a clean main

Not inherited from the earlier audit. The control is a worktree of `origin/main` with
`node_modules` symlinked and the WASM artefact copied — `engine/` and `package-lock.json` are
identical between main and M2, so the only variable under test is the web source. That mattered:
M2 touched `nodes-instanced.ts` and `node-scale.ts` in `a47d6208`, which is exactly what 3D
selection and `viewport-perf` exercise.

| Failure | Branch | Clean `main` | Failure text | Class |
|---|---|---|---|---|
| `basic-selection-permutations` 2D | fails | **fails** | `armed kinds for elements+nodes`, line 108 | preexisting defect |
| `basic-selection-permutations` 3D | fails | **fails** | identical to 2D | preexisting defect |
| `rc-design-visual` overlay legend | fails | **fails** | `696px → 697px, 645 pixels, ratio 0.03` | preexisting, non-blocking by design |
| `viewport-perf` `@perf` | fails on two models | **fails** on a third | `keyboard.up: Test timeout of 60000ms exceeded`, `orbitByKeyboard:211` | environment / harness |
| `basic-demos:160` | flaky | flaky | synthetic click on a moving target | declared flakiness |

The first three are identical character for character on both trees. **`viewport-perf` is not a
threshold failure**, which is what the earlier audit assumed: it wedges in `keyboard.up`, which is
why a different model fails each run — whichever one is running when it hangs.

`basic-demos:160` needed both conditions measured, and the isolated one was misleading:

| Condition | Branch | Clean `main` |
|---|---|---|
| Isolated, `--repeat-each=10`, first-attempt failures | **0 / 10** | **1 / 10** |
| In context, whole file `--repeat-each=3`, first-attempt failures | **3 / 3** | **3 / 3** |
| Retries consumed | `#1` every time | `#1` twice, **`#2` once** |
| Failing attempt / passing retry | ~24.7 s / ~10.6 s | ~24.7 s / ~9.5 s |

So it is deterministic and session-dependent rather than a low-rate flake: alone it passes, run
after its siblings in the single browser session `workers: 1` imposes it fails the first attempt
every time. `--retries=0` cannot suppress it — the spec declares
`test.describe.configure({ retries: 2 })` itself. No timeout was relaxed and no snapshot updated.

## CI

| Run | SHA | Contents | `e2e` |
|---|---|---|---|
| `33008435572` | `4b0afd2b` | all three decisions | **success** |
| `33008828740` attempt 1 | `8e6cc467` | the same plus one markdown file | **failure** |
| `33008828740` attempt 2 | `8e6cc467` | identical, rerun | **success** |

The diff between the two SHAs is one documentation file that enters no bundle, and attempt 1 failed
on `basic-demos:285` — the same test that made #156 red on its duplicated sha. Together with #156's
own pair, that is three independent demonstrations that this suite produces both colours on
identical code. `gh pr checks` shows only the last attempt.

Every area the three corrections could have touched is green in the full local suite: section
selector 67/67, generators 55/55, materials 16/16, composition and battens 24/24, cold-formed C/Z
15/15, joints 49/49, connections 66/66, metallic workflow 58/58. The four failures under "3D" are
`basic-selection-permutations` and `viewport-perf`; the metallic 3D specs pass.

---

# Why #156 looked red — diagnosed, and green again

Recorded here because the audit covering both branches lives on this PR, and because committing
anything to M1 would have moved it off the sha that just went green.

`gh pr checks 156` reported all seven green and exited 0 while the UI showed the PR as not
passing. Both were true. M1's head sha carries **fourteen** check runs across **two**
github-actions check suites; exactly one was red — `e2e` from run `32921289231` — and
`gh pr checks` prints one row per check name showing the latest, so it never appeared there. M2's
head has a single suite. That, not the code, was the whole difference between the two PRs.

The PR timeline dates the duplicate: `base_ref_changed` at 02:03:28 — #156 retargeted from
`feat/pro-steel-family` to `main` — then two `pull_request` runs created at 02:03:30 and 02:03:31
on the same head sha. The base change was legitimate. What turned it into two complete 28-minute
runs is that `ci.yml` declares no `concurrency`, so neither cancelled the other and one hit the
flake.

A third suite from the `render` app sits `queued` with zero check runs on M1, M2 **and main**
alike. It explains nothing; noted so it is not chased.

| Finding | Class |
|---|---|
| red `e2e` (`basic-demos:285`) | **flakiness**, platform-dependent — CI runner only |
| two runs on one sha | **misconfigured job** — no `concurrency`, triggered by a legitimate base change |
| `gh pr checks` green with a live red | **misconfigured job** — deduplicates by name, shows the latest |
| `render` suite queued | **environment** — dormant integration, same on main |
| `mergeStateStatus: BLOCKED` | **branch protection** — the missing approval; the ruleset declares **no required status check** |
| base / conflicts | none — `merge-tree` clean against main and against M2 |
| M1's own code | none — 74 files differ from main, **zero** of them Basic-mode |

Reproduced on a worktree of M1's exact sha: `basic-demos:285` passes 3/3 in the repeated file, and
the full `--grep @smoke` suite — precisely what CI runs — comes out **336 passed, 0 failed, 1
flaky, 4 skipped**, the same shape as CI's green run. It does not reproduce on darwin. The
mechanism is in the timing: CI takes 25.4 minutes for what takes 10.5 locally, and the test throws
four synthetic canvas clicks hoping one lands on a member.

The asymmetry that lets this one go red rather than flaky: `:160` declares
`test.describe.configure({ retries: 2 })` for exactly this reason and `:285`, structurally the same
walkthrough, has none. Not fixed here — `basic-demos.spec.ts` is byte-identical between M1 and
main, and changing it from a steel branch would give M1 ownership of a Basic-mode spec it has no
reason to touch.

CI was rerun on the same sha with nothing changed. Attempt 2 of `32921289231` is green; all
fourteen check runs are now success or skipped, both suites conclude success, and #156 stopped
showing red without a line moving. `concurrency` was deliberately **not** added to `ci.yml` — it is
main's infrastructure, and `cancel-in-progress` trades a red for a cancelled run unless the group
is chosen deliberately. The patch is written down in §30 of
`docs/handoffs/m1-m2-ci-audit-and-three-decisions.md` for whoever owns it.

---

# The two findings that were left open, audited to their cause

Part III of `docs/handoffs/m1-m2-ci-audit-and-three-decisions.md` left `.conn-ratio-badge` and the
`svelte-check` silence open, with evidence and no diagnosis. Part VI closes both. Neither is what
the earlier reading said it was.

**`.conn-ratio-badge` is dead CSS, and M1 is what killed it.** The census is 0 applications of the
class in `main`, in M1, in the merge and here — all four occurrences sit inside `<style>`. The
earlier account blamed main's `2c79ed52` and the union merge, which explains how the rules came
*back*, not how they died: `git log -S 'class="conn-ratio-badge'` puts that on **`b71432cd`**, M1's
own sub-section split, two days earlier, which moved the governing ratio to `StageSection`'s neutral
`badge` prop and deleted both spans along with `statusClass()`. `851fd57b` then measured the
contrast of those rules and rewrote all three states, and five assertions pinned the result — a
commit and a test suite about CSS the browser never painted. So there is no product disagreement to
settle. Removed in `6448e89d`, on M1, in a commit of its own, and merged down here; the ratio itself
is untouched and still printed by both sub-sections. Three hand-written guards replace the five
assertions: no `conn-ratio-badge` in the stylesheet, the percentage still reaches `StageSection`,
and no element on the panel carries `st-ok` / `st-warn` / `st-fail` — so a ratio gets no status hue
at all and the auxiliary block's own `within / near the limit / over the limit` stays the only thing
entitled to comment on it. No green tick and no normative verdict came back.

**`svelte-check` does read `ProConnectionsTab.svelte`.** An injected type error is reported at
`2:9` in the same run where an injected dead selector is not. The blind spot is upstream of the
checker: Svelte's `css-prune` stops pruning as soon as a component has a class attribute it cannot
read statically, and this one has four (`class="conn-result-card {auxTone(...)}"`). Minimal repro:
`<div class="a">` warns about an unused `.zz-dead`, `<div class="a {x}">` does not. Measured over the
whole tree by appending a dead selector to every styled component and compiling: **30 of 169 are
blind the same way, 24 of them under `components/pro/`**. The 116 `Unused CSS selector` warnings in
the same run are the mechanism working where it can.

**No CI job was added, and the reason is not cost.** A `svelte-check` job over `9883e2bd` would have
printed exactly the 0 mentions the manual run did — it would not have found the first item. The
separate gap from Part I §2 is real and is now costed rather than acted on: `check:gate` runs
end-to-end in **14.4 s**, every metallic path is at **0 errors**, so extending `GUARDED_PATHS` and
adding one step to the existing `web` job is free. Proposed with files, impact and tests in §32 and
left for whoever owns `ci.yml`, on the same grounds as the `concurrency` patch in §30. What no job
fixes, and what is documented as a standing limitation: dead CSS on those 30 components is findable
only by a manual census, or by an assertion written by hand.

---

# Part VII — the designed joints reach a share link (I-08), and A-01 becomes a decision

## The base moved

`origin/main` (`b579de87`) went into M1 by explicit merge, and M1 went into this branch the same way. No rebase, no force-push, no conflicts either time. The two branches now share the same base of `main`, and this PR stays based on `feat/pro-steel-m1`.

M1 also carries one E2E fix that this branch inherits: `basic-selection-permutations.spec.ts` read `aria-checked` where `SelectionPanel` publishes `aria-pressed`, so its fifteen-permutation audit had been failing on the first two-kind subset from the start — and had never run anywhere, because CI's `e2e` job runs `--grep @smoke` and that file carries no tag. Details in #156.

## I-08, the joints half — closed

Part VI put the joint choices on `StructureModel.jointDesigns`, so they travel `.ded`, undo/redo, tab capture and autosave. The URL codec was not one of those paths: it is a separate versioned wire format that enumerates what it carries. A shed with twenty designed joints, shared by link, arrived with twenty undesigned ones — and the worst of the three surfaces that build such a link is the feedback widget, which attaches one to every report. A user reporting a problem with a joint sent a link that rebuilt the model without it.

### The format

One new top-level key of the compact object, and **no position added to any existing tuple**:

```
jd: { v: 1, j: [ [ nodeId, memberCount, [x,y,z] | 0, choices ], … ] }
```

Four fixed slots per joint; all growth room is inside `choices`, which holds four groups (`b`, `p`, `w`, `ba`) of one- and two-letter keys. The length discipline is the section tuple's, because `MAX_URL_SAFE` is a limit `ToolbarProject` actually checks.

`SHARE_VERSION` 4 → 5, compatible in both directions:

- **old link, this build** — no `jd`, which reads as "no joint decisions". Not an empty joint designed at every node.
- **new link, older build** — the two existing migrations key off `sv >= 3` and `sv >= 4`, which 5 satisfies, and a reader that predates `jd` ignores a key it does not know.
- **later link, this build** — unknown keys inside `choices` are ignored, so a field added later as an optional key does not make today's build reject tomorrow's link.

### Only the choices, and only what can be vouched for

`packJointDesigns` walks a field table and emits nothing else. A `capacityKN`, a `utilisation` or a `checks` array that reached the model through a hand-edited file **cannot** reach a URL through it — asserted by tampering with a model on purpose and grepping the wire. Capacities are recomputed from the open model on every read.

Obsolescence is **not** carried. `nodeMissing`/`nodeMoved`/`topologyChanged` are decided against the model that is open, on every read; they are derived, so they are not part of the persisted decision. The fingerprint they are derived **from** is — `atMm` and `memberCount` — because that records the joint the user was actually looking at.

Carrying it is what extends I-07 to the link. A link opened over a different model reports its joints **obsolete**, with a reason and two remedies, instead of matching them by node id and presenting as chosen something nobody chose there. A payload that arrives with **no** fingerprint reconciles as `nodeMoved` — never as a match.

### Refused rather than half believed

A known key with the wrong type, an entry that is not a well-formed tuple, or a container version this build cannot read all throw, and `decompressV2` turns that into the `null` its callers already handle.

Rejecting the whole link rather than dropping the joints is deliberate, and the cost is written down rather than hidden: the model bytes and the joint bytes arrive in **one** deflate stream, and treating half of a stream that is not what it claims to be as trustworthy is the weaker position. What it costs a recipient is stated in `share-codec-fields.md` §7.3, including that the `#data=` path answers a rejected link by opening empty — the same silence any corrupt link already gets, and not this change's to redesign.

### The drift guard — the piece the handoff asked for in §6 and never had

Four section fields went missing from this codec without anyone noticing, because nothing failed when a field was added and the codec did not hear about it. Two halves now, and **both were verified by breaking them on purpose**:

- drop a field from the codec's table → the completeness test fails, by name;
- add a field to `JointChoices` → `npm run typecheck` reports exactly **one new error**, pointing at the `Required<…>` fixture that no longer lists every key.

The permitted-value lists are built from `Record<Union, 1>`, so adding a member to `ThreadCondition`, `EdgeFinish`, `Exposure`, `WeldProcess` or `WeldLoading` does not compile until the wire accepts it. `BOLT_GRADES` and `BUILT_UP_ARRANGEMENTS` are imported: Tabla J.3.2 and the seven arrangements have one list each, in one file each. The same pattern is written up for `interface Section` to copy.

### Measured, not assumed

Twenty fully designed joints — bolts, plate and weld on every one — cost **391 characters, 19.6 each**: the first 203 and each one after it about 10, because the key names repeat identically and that is what deflate collapses. The test's budget is 30, half again over the measurement.

### Tests

**55 new**, against the code that runs. The codec is its own pure module precisely so they can be: `url-sharing.ts` imports four stores, which is why `url-sharing.test.ts` tests an **inlined copy** of the encoder, and a wire format tested through a duplicate is a wire format with no test.

Round-trips for a model with no joints, a bolted joint, a weld, battens, several nodes and every bolt grade the strength table publishes; absent optional fields staying absent rather than arriving as zero; an explicit `null` told apart from a key never set; nineteen corrupt payloads refused; exact reconstruction field for field; nothing computed on the wire; a joint refusing to follow a link into another model, with each of the three obsolescence reasons; an old `sv:4` link still opening and opening with no joint decisions; the M1↔M2 direction asserted as the property that makes it true — exactly one new top-level key, everything else byte-identical; and the length budget.

**The four section fields of I-08 are untouched and still open.** They are preexisting, they belong to concrete and steel alike, and `built-section-contract.test.ts` still asserts their loss.

## A-01 — explained, not decided

A-01 is the one inventory item that is a product decision rather than a defect, and the inventory ranks it third by priority: not because anything is broken, but because reverting it after a merge costs far more than reverting it now.

It is written up in product language in `docs/handoffs/a-01-decision.md` — what the user sees now against what they saw on `main`, which of the three withdrawn ways in was equivalent and which was not, whether data is lost, whether QA sees it, whether it blocks the merge, the minimum correction, and four alternatives with what each costs. **Nothing was implemented**: it changes visible behaviour, so it stays a pending decision.

The findings, since they are the part that decides it:

- **No data loss.** No format changed; no saved project reads differently. The prerequisite the decision exposed — a section built *through the dialog* dropped `tw`, `tf`, `t` and `tl` — was fixed before anything was deleted.
- **The inline profile table was not equivalent** to what survived: it called `addSection` directly, so a section entered with no arrangement, no gap, no rotation and no `ProfileSpec`; its rows were `<tr onclick>`, unreachable by keyboard; and it offered no standards body, design code or depth filter, so the catalogue looked smaller than it is.
- **The inline build form was equivalent**, and the survivor is the better of the two: same lists, same computed properties, same `built` record, plus an `initial` that lets a built section be reopened and edited, plus the only tests of the pair.
- So what changed for a user is **one extra click** to reach the builder, and a habit break with no signpost. That is the whole of the divergence from the written proposal, which said the builder was not to be touched.

## Gates

typecheck **473 = baseline 473, no new errors** · unit **8103 pass / 0 fail** (437 files, 12 skipped, 1 todo) · build project **19 pass** · `npm run build` ✓ · i18n parity ✓ (`locale-parity`, `steel-keys`, `pro-flow-coverage`) · full local E2E on its own `E2E_PORT`.

No Rust, Cargo, WASM or solver change. No snapshot updated and no timeout inflated. No new PR opened. Both PRs stay draft.

---

# Review summary — scope, limits, and what is deferred

## Delivered scope

Everything in Parts I–VII above. In one list:

- **Section and material selectors** behind a single modal, with the deep grade catalogue, the provenance card and the keyboard path — one way in, not three (A-01, below).
- **Generators** and the metallic workflow, with the capability gaps stated on screen rather than implied.
- **Joint design** — bolts, plate, weld, battens — with its own verdict vocabulary (`within / near the limit / over the limit`), deliberately distinct from the canonical one.
- **I-06** — joint choices live on `StructureModel.jointDesigns`, so they travel `.ded`, undo/redo, tab capture and autosave.
- **I-07** — a stored joint records what it was designed against, and one that no longer matches is *obsolete*: kept, shown with its reason, never applied.
- **I-08, the joints half** — the designed joints reach a share link, `SHARE_VERSION = 5`, choices only, with the drift guard the handoff asked for and never had.
- **B-01** — the modal is the only way to add a material, with a custom-material form bounded on physics rather than on a catalogue.

## Merge order, and the dependency — read this first

**`#156` merges into `main` first. `#164` merges second.**

This branch is based on `feat/pro-steel-m1` and **contains it by merge**. It is not independent of #156 and must not be retargeted at `main` to merge earlier: doing so would carry M1's entire diff in under this PR's review. The stack is M1 → M2, in that order, and the base field of this PR is the record of it.

## Known limitations

| Limitation | Where it is written down |
|---|---|
| **No metallic export, no metallic verification.** No clause map, no joint state in any export; `steelCountsAsVerified()` returns the literal `false` | I-09 of the QA inventory |
| **The share codec still drops four section fields** — `composition`, `profileFamily`, `tl`, `built`. Preexisting, concrete and steel alike. The joints half is done; this half is not | `share-codec-fields.md` §1–§6; asserted as a declared loss by `built-section-contract.test.ts` |
| **A corrupt joints payload costs the whole link**, deliberately — the model bytes and the joint bytes arrive in one deflate stream. By the `#data=` path a rejected link opens empty without saying why, which is the silence any corrupt link already gets | `share-codec-fields.md` §7.3 |
| **Obsolete joints are excluded from `designedNodeIds`**, so a document never tabulates one | I-07 |
| **No retroactive provenance.** A project saved before these branches does not say by which route each section entered, and nothing is invented to fill it | I-11 |
| **Pinned lattice bases carry no out-of-plane restraint**, and say so | `generator.assume.latticeBasesPinnedNoOutOfPlane` |
| **`docs/handoffs/m1-m2-qa-manual-inventory.md` is the inventory as of the start of review**, not as of this head. Six entries were re-measured after it was written; the corrected file is on `backup/qa-inventory-remeasure-2026-08-27` and is deferred to M3 (see the M3 table). Reading an entry that contradicts the code means reading the snapshot, not a defect in the code | This PR's own commits and the other handoffs are current; only that one file lags |

## A-01 — the one open decision, and it is not a blocker

`docs/handoffs/a-01-decision.md` states it in product terms. Summary of the findings, because they are what decides it:

- **No data loss.** No format changed, no saved project reads differently. The prerequisite the decision exposed — a section built *through the dialog* dropped `tw`, `tf`, `t` and `tl` — was fixed **before** anything was deleted.
- **The inline profile table was not equivalent** to what survived: it called `addSection` directly, so a section entered with no arrangement, no gap, no rotation and no `ProfileSpec`; its rows were `<tr onclick>`, unreachable by keyboard; and it offered no standards body, design code or depth filter.
- **The inline build form was equivalent**, and the survivor is the better of the two — it accepts `initial`, so a built section can be reopened and edited, and it is the only one of the pair with tests.
- So what changed for a user is **one extra click** to reach the builder, plus a habit break with no signpost.

**Classified as apt for M3, with one caveat stated rather than buried:** it is not a correctness error, not data loss, not a false claim and not a broken flow — so it does not block. But the cost of changing it *grows* after merge, because the tab went from 724 lines to 205 and three regression tests now assert the absence of the machinery. Deciding it before merge is cheaper than deciding it after; not deciding it does not make the merge unsafe.

## Non-blocking findings, for M3

| Finding | Why it is not blocking |
|---|---|
| **A-01** — see above | Product decision; nothing false or lossy |
| **The QA inventory in this branch is a snapshot, and six of its entries have since been re-measured.** A corrected version exists — **`docs/handoffs/m1-m2-qa-manual-inventory.md` on the branch `backup/qa-inventory-remeasure-2026-08-27`**, 3464 lines against this branch's 3205. Four entries in the shipped version state the OPPOSITE of what this branch does (B-01, I-06, I-07, I-08), plus B-13, and the corrected version adds **E-25**, the obsolete-joints notice — the visible half of I-07's repair, which did not exist when the inventory was written | **Deliberately kept out of this PR**, by the branch owner's decision: the commits appeared in the worktree without attributable authorship after this PR was already marked ready, so the head under review stays as declared. The work is preserved on that branch and is an **input to M3**, not a loss. Its own "blocker 7" is the one item now stale — it says this PR's `e2e` is red, and that fell when the job was relaunched and passed |
| **The four section fields of I-08** | Preexisting, not steel-only, and the loss is asserted rather than hidden |
| **The URL container version is the same constant the `.ded` writes.** Whoever next increments `JOINT_DESIGNS_SCHEMA_VERSION` for a `.ded` reason makes every already-shared link stop opening **entirely** — not "without its joints" — because §7.3 rejects the whole link on a version it cannot read | Unreachable today, with one version in circulation, so it is a trap for the next change rather than a defect now. Both ways out, and why neither is mechanical, are written into `share-codec-fields.md` §7.1 |
| **The `basic-demos` section-walkthrough flake** — narrowed to two candidate causes, not diagnosed. `main`'s `0844f01d` already carries containment (declared retries) and calls itself *"containment, not a diagnosis"* | Not this branch's surface. That containment is **not yet in this branch** — it arrives with the next `main` merge — so the test can still redden this job until then |
| **`returns a mechanism wearing a number under a load along the building` is 2.89 s** against the 5 s default | Nothing to split: a single solve, slow because the unbraced shed is near-singular *by design*, which is the assertion itself |
| **CI gaps** — no `typecheck` job, no `svelte-check` job, no `concurrency`, and `e2e` runs `--grep @smoke` only | Owned by whoever owns `ci.yml`. The `@smoke` gap is what hid a spec that had been broken from the start |
| **Dead CSS is invisible on 30 of 169 styled components** — Svelte's `css-prune` stops pruning once a component has a class attribute it cannot read statically | Measured and documented; findable only by a manual census |
| **A darwin screenshot baseline is 1 px wider** than the committed one (697 vs 696, 3 % over a 2 % tolerance) | Fixing it means updating a snapshot; CI runs that job only on `main` or under `run-e2e` |
| **`Install wasm-pack` reports success when its download fails.** The step is `curl … | sh`, so its exit status is `sh`'s — a failed `curl` leaves the step green and the failure resurfaces two steps later as `wasm-pack: command not found`, exit 127, with the blame on `Build WASM engine`. Observed on this PR: `curl: (35) Recv failure: Connection reset by peer` | Transient network fault, not a branch defect — a re-run clears it. The durable fix (`set -o pipefail`, or download-then-run, or a pinned release) is in `ci.yml`, owned by whoever owns `main` |
| **`@perf` viewport measurements time out under whole-suite contention**, 4/4 in isolation | Fixing it means inflating a timeout; CI never runs `@perf` |

## Nothing classified as blocking

No correctness error, no data loss, no broken main flow, and no false claim left standing. One item did fall in the false-claim category and is retracted in #156: I published a diagnosis of the walkthrough flake that said the test skips opening the Section Analysis panel. It is wrong — `stressQuery` is written by the viewport click handler when `findNearestElement(…, 0.3)` finds a member, and the panel opens as a *consequence* of that write, not as a precondition. The finding is narrowed, not diagnosed.

## Gates on this head

| Gate | Result |
|---|---|
| `typecheck` | **473 = baseline 473**, no new errors |
| `test:unit` | **8104 pass / 0 fail** (437 files, 12 skipped, 1 todo) |
| `test:build` | **19 pass** |
| `npm run build` | ✓ |
| i18n parity | ✓ `locale-parity`, `steel-keys`, `pro-flow-coverage` |
| Full local E2E | **826 pass** — 2 failed, both in the table above and neither fixable inside the stated constraints |
| CI | `lint` · `test` · `suite (1)` · `suite (2)` · `web` · `e2e` |
| Manual server | `http://127.0.0.1:4004`, HTTP 200, serving this branch with M1 + M2 |

`main` has moved on since M1 synced with it (`b579de87` → `c4540883`). Deliberately **not** chased — the branch is frozen for review. A dry-run merge of the new `main` into M1 produces **zero conflict markers**; the single file both sides touched, `basic-demos.spec.ts`, is changed in different regions.

No Rust, Cargo, WASM or solver change. No snapshot updated, no timeout inflated, no new PR opened. Both PRs stay draft.


diegokingston:
## Review

Context notes first: M1 is merged, so the `main` base is now right and the "do not merge before M1" box in the body is stale. CI is green except the e2e smoke failure on `basic-demos.spec.ts:320` ("the section walkthrough › advances when the reader clicks the member") — the same known flake this PR documents and adds `retries: 2` to. It failed all three attempts in this run; per the PR's own comment ("if the retries stop absorbing it, that is the signal to do that work"), the real repair (ask the app where the member is instead of guessing) is now due.

The bulk of the diff was reviewed for numerical correctness. What's clean, briefly: `connection/` (fillet-weld §J.2.2/J.2.4, bolted-joint Tabla J.3.2 values verified against the shipped CIRSOC markdown, §J.3.6/J.3.7/J.3.10, bolt-geometry J.3.3/J.3.4/J.3.5, batten §E.6) is sound and hand-computed-test-backed; built-up section composition applies Steiner with correct Iyz sign handling; persistence/share-links are versioned and strictly validated; i18n parity holds; the sampled e2e specs assert real state machines and clause references.

### Blocker candidate — steel flexure pairs each axis's capacity with the other axis's demand

`web/src/lib/engine/verification-service.ts:497-501` swaps the inertia mapping (`Iz ← section.iy`, `Iy ← section.iz`) — the swap itself is correct and fixes a real bug (`ry` from the strong axis inflated `Lp` by ~√13.7 ≈ 3.7×, judging LTB members as inside the plateau). But the demand mapping at `:535` still sends `Muz ← mz` (and `Muy ← my`), and the checker's contract is `Muz` = strong-axis moment (`cirsoc301.ts:103`), with LTB running on `Muz` against the strong-axis capacity derived from `Iz`.

The app's own convention says gravity bending of a horizontal beam is **My**, not Mz:

- `web/src/lib/engine/types-3d.ts:351-359`: "Mz — moment about local z; bends over the section width (uses Iz)… For the default unrolled tall section **My is the strong/vertical axis**".
- `web/src/lib/engine/auto-verify.ts:221-225` (the RC path): "a horizontal beam's vertical-plane (gravity) bending is **My** (depth = h, paired shear Vz)… beams were stuck at Mz → 0" — RC needed a per-element axis-aware branch for exactly this; the steel path never got one.
- `engine/src/element/transform.rs:177-186` and the stiffness blocks in `frame.rs`: solver `iy` pairs with output `my`, solver `iz` with `mz`; and `solver-service.ts:1446-1453` maps `section.iy → solver iy`. So `my ↔ section.iy` (strong), `mz ↔ section.iz` (weak).

Consequence as shipped: a gravity-loaded beam's real moment (`my`) is checked as *weak-axis* flexure against `φ·Fy·Zy` (~4× below strong-axis for an IPE — false failures), while the strong-axis LTB check runs on `mz ≈ 0` and passes vacuously (unconservative). The inertia swap and the demand mapping contradict each other; one of them has to change, and `steel-axis-mapping.test.ts` can't see it because it feeds `mz` as a fixture rather than solving a beam. This needs a solved-beam assertion (simple/cantilever beam through the real solver, utilization vs hand calc), not another fixture.

### Should-fix

2. **`Cb` is only applied in LTB zone 3 (elastic), not zone 2 (inelastic)** — `cirsoc301.ts:496-504` (`Mn = Mp − (Mp−Mr)·(Lb−Lp)/(Lr−Lp)`, no `Cb`) vs `:510-538` (zone 3 uses `params.Cb ?? 1.0`). CIRSOC F.2.2 is `Mn = Cb·[Mp − …] ≤ Mp` — the factor belongs to the inelastic branch, which is where most real members land. As shipped the headline Cb feature is inert for them, and `steps` prints "Cb = x (F.1.1, calculado del diagrama)" unconditionally (`:452-454`) even when the value was never used.
3. **Two divergent bolt/weld implementations are both user-reachable** — `engine/connection-design.ts:62-135` (used by `ProConnectionsTab.svelte:335-355`) vs the new `connection/`. Differences that change numbers: elliptic `(fv/Fv)²+(ft/Ft)²` interaction instead of §J.3.7's linear reduction; standard hole `d+2` while new Tabla J.3.3 gives `d+3` for d=24/27; old weld `β=0.6` flat above L/w=300 (unconservative vs the new 180·w cap); no 1.5/3.0 bearing pair. At minimum cross-label which is authoritative; ideally retire the old path.
4. **Tabla J.3.2 note (e) not implemented nor named** — `connection/bolted-joint.ts`. The 20% reduction for splices with bolts spaced >1000 mm parallel to the force is in neither the covered nor the refused list. Silent unconservatism for long splices.

### Nits

5. `moment-gradient.ts:165-170` — comment claims F.1.1's own maximum is 3,0; the expression's bound is 12,5/2,5 = 5,0. Comment-only, but it's the stated reason no cap is applied. Related: `verification-service.ts:472-480` feeds `momentGradient` a per-station cross-combo envelope of |mz|, which is not the per-load-case diagram F.1.1 describes — typically conservative, not provably so.
6. `missingSteelInputs` (`verification-service.ts:303-315`) requires `tw`/`tf` for all steel, so CHS/RHS tubes (which carry `t`) are always skipped with gaps labelled `webThickness`/`flangeThickness` — a misleading remedy for a section the I-based checker legitimately can't take.
7. `joint-design.ts:60-63` — `SEVERITY` ranks `notDesigned` above `incomplete`, so a fully designed bolt group plus a blank opened weld panel downgrades the joint to `notDesigned`. Arguable; worth a deliberate decision.

PR #166, opened 2026-08-22
mobile: one shell for Basic and PRO on a phone

Batuis:
Basic and PRO each had two interfaces: one for a desktop and a different, smaller one for a phone. This makes each of them one interface that adapts.

## The problem

Everything built for Basic since the ribbon landed — the pointer mode on the model, the Selection panel, the results selectors, the colour-scale switch, the eight walkthroughs — was reachable only through the ribbon, and the phone mounted no ribbon. It got the old left `Toolbar` in a drawer plus a floating tool strip instead. Any new work on Basic had to be done twice or it silently shipped to desktop alone.

PRO's phone shell was more deliberate — a five-button bar and a native `<select>` listing thirteen panel tabs — but it could not reach the four stages, the eight diagrams or the two colour maps at all, because those are not tabs.

## Basic

The ribbon mounts at every width and degrades instead of forking. Below 768 px it is nine controls sharing the width — Project, undo, redo, Selection, 2D/3D, Model, Solve, Advanced, Results — with the group captions giving way to the rules that already separate the groups.

`Model` opens the Model-data panel rather than a menu, because that panel's six tabs already *were* those six buttons: `pickTab` arms each tab's tool, and has since it was written. On a phone the tab strip becomes one row of six icon targets, pinned above the table.

`BasicPanel` is the bottom sheet, and it **shares** the screen rather than covering it: `.app-body` reserves `--st-sheet-h`, so the canvas is the size it appears to be. Without that, choosing a moment diagram drew it behind the panel that had just opened to control it. A grab handle drags the sheet between 22 % and 86 %; the handle is the only surface that resizes and the body scrolls as it always did.

Removed from Basic, all of it a second route to what the ribbon carries: `FloatingTools`, both mobile drawers, the mobile bottom bar and `MobileResultsPanel`.

## PRO

Four stages and twenty-five commands do not fit a touch row — ANALYSE alone carries fifteen. So the bar keeps only verbs, and both halves of "where am I" are two pills in the sheet: command on the left, stage on the right. Project is an entry in the left pill and greys the right one, because the document belongs to no stage.

The nine camera buttons that ran down the right edge of the model become one split button: the face is whichever control you used last, the caret opens the rest with their names.

**`src/lib/pro/stages.ts` is what makes this survive PRO growing.** The command tree was inside `ProRibbon.svelte`; two surfaces draw it now, and a definition owned by one is a definition the other forgets. A command added there gets a ribbon slot and a grid cell, and nothing else has to be revisited. `src/lib/pro/camera-actions.ts` does the same for the view controls.

`stages-coherence.test.ts` guards it. The realistic failure is not a careless edit — it is someone on another branch adding a good command without knowing `PRO_TAB_STAGE` is a second list that has to learn about it: the ribbon looks right and the phone silently names the wrong stage, because the lookup falls back to `'model'`. Verified by breaking it.

## Defects found and fixed along the way

- **`Toolbar` was mounted twice on a phone** — in the drawer, and again in `.app-body` behind `leftSidebarOpen` where `.sidebar { display: none }` hid it. Two live copies of a 2,400-line component that was never shown, with every id inside it duplicated.
- **PRO's Calcular was wired to nothing.** `proPanelRef` was bound only to the desktop `ProPanel`, so on a phone it was null: the button read `canSolve()` off nothing and rendered permanently disabled. Both instances are bound now.
- **The restore-a-project prompt was 19 px wide** on every phone, in both modes. It renders inline in a header that holds a logo, a mode selector, a project name and a settings button in 375 px, and got what was left. Present in the DOM, so every check asking whether it exists passed. It is drawn over the top of the model now.
- **The toast container swallowed every touch.** The phone rule set a `top` without clearing the desktop `bottom`, so a fixed invisible box spanned the viewport at z-index 1100 for as long as each message lived — after every solve, the canvas could not be panned or drawn on.
- **Settings had no opener on a phone**, in either mode, while the language selector had already been moved into it *for that width*. The control was hidden from where it was and hosted where nothing could open it.
- A `basico && isMobile` branch sat inside a `!isMobile` block, unreachable since it was written.

## The walkthroughs work on a phone

They never had — until this branch the phone had no ribbon to point at. Five anchors did not resolve at 375 px and needed two different repairs: `sections` and `materials` are data-panel tabs there, and the diagrams live inside a menu the step now opens first.

Verified by walking every step at **both** widths and diffing. A step that lights its target on a desktop and not on a phone is the defect, and neither run alone can show it — a phone-only run cannot tell "this step lost its target" from "this step never had one".

## Verification

- `npm run typecheck` — 479, unchanged from baseline
- `npm run test` — 379 files pass. `chs-shear-agreement` fails, unchanged from this branch's base and untouched here (circular tube shear, #157)
- `npx playwright test` — full suite
- The walkthrough audit reports `sin problemas` at 1500 × 950 **and** at 375 × 667. It takes a viewport now; the second run is new and is the one that means anything about a phone
- Every visible, enabled control in both phone shells pressed with app state read before and after: **zero dead controls**

## Known and deliberate

- Nine controls at 44 px come to 396 px, so Basic's row gives each 39–41 px of width at 375. Always visible beats correctly sized and off screen; the height stays 44
- `pointer-mode` and `zoom-to-fit` remain 32 × 32. They are a documented pair on the canvas and have to move together, where 44 px costs model area
- Control size in the panels is a setting (Ajustes → Móvil), defaulting to compact. 44 px everywhere adds 124 px of scrolling before the results table, and neither side wins outright on a 375 px screen

`docs/handoffs/mobile-ui-pr166.md` carries the measurements, the traps and what is left.


diegokingston:
## Review

Overall: solid. The claimed fixes check out — Toolbar is genuinely mounted once now, `proPanelRef` is bound on both ProPanel instances (and the `tick()` before `solve()` is correct), the toast container no longer swallows touches, the `--st-sheet-h` reservation + ResizeObservers actually re-frame the canvas, i18n parity holds across es/en/pt, and `stages-coherence.test.ts` covers the failure mode it documents (broke it mentally; the `?? 'model'` fallback at `ProPanel.svelte:812`/`ProRibbon.svelte:154` is what it guards). The PRO mobile e2e asserts a real solve (`solveCount` increments), not just smoke.

### Should-fix

1. **`MobileResultsPanel` is a dead-end on PRO mobile.** The panel is still mounted (`App.svelte:1481`) and still opens automatically on any mobile solve (`web/src/lib/actions/solve.ts:68,119`, `web/src/lib/engine/live-calc.ts:216,252,347`), but its manual opener (`.pmt-results`) was removed. So after Calcular the floating panel pops up on top of the results sheet — the two-panels-arguing situation this PR removed from Basic — and once closed with ×, nothing can reopen it short of re-solving. Either wire an opener into the new bar/sheet, or stop opening it from the solve paths (and drop the panel, since the sheet is now the results surface).
2. **Dead store writes.** `uiStore.leftDrawerOpen` is still written in `solve.ts:67,118` (the left drawer renders nowhere now) and throughout `tour/tour-steps.ts` (which appears to have no importers left — dead file?). `mobileResultsPanelOpen = true` also fires on Basic mobile solves where the panel never renders (`MobileResultsPanel.svelte:47` requires `appMode === 'pro'`).
3. **Stale comments** describing controls that no longer exist: `App.svelte:1247` (stage menu `top: 100%` of `.pmt-wrap` — the stage selector moved into the sheet) and `App.svelte:3064` (`.pmt-btn.active` — "Project and the stage are the only two…", but the bar has neither any more).

### Nits

- `SheetGrab.svelte:88-101`: window-level `pointermove/up/cancel` listeners are only removed in `up()` — unmounting mid-drag leaks them until the next pointerup (self-healing, but an `onDestroy` teardown is one line), and a second `pointerdown` (multitouch) installs a second listener set with stale `startY/startVh`.
- `SheetGrab`'s `onResizeEnd` prop has no callers (`App.svelte:1634`, `BasicPanel.svelte:174`) — dead optional API.
- PRO's sheet doesn't re-frame on open/close (Basic does via the `sheetWasOpen` effect, `App.svelte:131-149`) — opening the PRO sheet shrinks the canvas 45vh with the old framing.
- Educación's `.drawer-right` still hardcodes `58vh` while `tokens.css:170` documents 45vh for the sheet — the two surfaces disagree.
- Dead CSS: `.rb-cmd.prominent` rules in `Ribbon.svelte` (the sole `prominent` command was removed); the `ribbon.groupData` locale key is now unused in es/en/pt.
- Tour: the `sections`/`materials` steps (`modelling-2d.ts:135,144`) resolve to `dt-tab-*` inside the data sheet but don't open the sheet in their own `onEnter` — they rely on an earlier step having left it open. Closing the sheet mid-tour loses the anchor.
- `web/scripts/audit-demos.mjs` only audits at 1500×950; the anchor repairs are specifically about 375px, which the audit never exercises (the PR body says the audit takes a viewport now — the script in this diff doesn't seem to).

PR #170, opened 2026-08-24
H2 — PRO concrete flow: stage timeline, design/detailing redesign, documents, raw-forces report

Batuis:
**Ready for review.** Continuation of
[#161](https://github.com/lambdaclass/stabileo/pull/161) (H1 — Concrete PRO), stacked on
`feat/pro-concrete-h1` and containing it by merge. This diff is H2's own 66 commits:
**183 files, +29 038 / −2 630**.

**Merge order is fixed: #161 (H1) first, then this one.** H2's base *is* H1, so landing this first
would either pull H1's 37 commits in unreviewed or force a rebase.

## Scope

The five-stage PRO concrete flow, from the stage strip down to the report.

| Phase | What landed |
|---|---|
| **F0** | Contracts, no visible change. Split `detailing.svelte.ts` (1566 LOC against an 800 ceiling) and prepared `ProPanel.svelte` (1316 against 600). Persistence approved and implemented in F0.6. |
| **F1** | The stage timeline. One vocabulary replacing three, a sticky strip, and CHECK no longer shown before DESIGN. |
| **F2** | The DISEÑAR stage. `ProDesignTab` moves inside the stage; the two design commands with different scopes reconciled. |
| **F3a–F3d** | Detailing: the member list, bar locking, editing, conflicts. |
| **F4** | Documents. |
| **F5** | The raw-forces report. |
| **F6** | The viewer: compact dismissible notices and a right-hand selection panel. |
| **F7** | Performance — ten operations measured before anything was optimised. |

The mandatory audit — three languages, 1280×720 and 1024×700, keyboard, focus, accessible names,
empty states, overflow, tokens — was the closing condition of **each** phase, not a final one.

## What the audit found before any code moved

**Three incompatible stage vocabularies, not one numbering bug.** `WorkflowStages` drew six
stages, `ProRcWorkflowTab` had five disclosures, `onGoTo` accepted a third set of five. `demands`
and `check` were drawn with no destination; `floors` was a destination with no stage; `model` and
`design` both fell through to the same `scrollTo`. The strip was also not sticky — the column is
`overflow-y: auto` and scrolled it away.

**And a semantic defect underneath it.** The strip put CHECK third and DESIGN fourth, so it read
"verification, then design". A completed *Verificación* shown before anything has been designed
says the reinforcement was verified. That is the one claim this branch may not make by accident.

**The 2D/3D sync needed no new mechanism.** `rebarWorkspace` already holds a single `selection`,
and its own comment says a second channel is how a panel comes to highlight one thing while the
viewport highlights another. All eight consumers were overlay components; `DetailingWorkflow` and
`ProRcWorkflowTab` referenced it zero times — which is precisely why the detailing list and the
viewer were two independent representations.

**The raw-forces report is backed by real solver output**, and better than first reported:
`buildCriticalStations` seeds 0, 0.25, 0.5, 0.75 and 1 *unconditionally*, so on a normal member
the quarter grid is a subset of what the engine already computed. It is not a subset for a
zero-length member, which short-circuits to `[0, 1]` — which is why the default is defined as an
evaluation rather than a filter.

## Also here: one test file, no product code

`f0700fb0` adds `engine/tests/paper_examples.rs::a3_building_column_removal` — the JAIE 2026 §6.3
column-removal case, which had no test behind it. Its header states line by line what comes from
the paper (topology, 20 nodes, 27 elements, 5 supports, ΣRz = 1500 kN) and what was chosen for the
reconstruction (spans, heights, sections, load), so nobody later reads the chosen values as
published ones. Removing the interior ground-floor column multiplies the governing beam's moment by
3.97 against the 3.54 the paper's own D/C pair implies; the sections were **not** tuned to close
that gap. It also cross-checks the boundary: the Rust solver called directly reproduces the moments
the WASM build produces in the browser.

## CI and gates

**GitHub CI on `ef79f8c2`: green.** `lint`, `test`, `suite (1)`, `suite (2)`, `web`, `e2e` all
pass; `bench` skipped by configuration.

This PR carries the **`run-e2e`** label, so that `e2e` job is not the usual smoke-only one — all
three Playwright passes ran on Linux and all three are green:

| CI step | Result |
|---|---|
| `E2E smoke suite` (blocking) | success |
| `E2E slow suite` — `--grep @slow --grep-invert "visual baselines"` | **success** |
| `E2E visual baselines` (`continue-on-error` by decision O7) | **success** |

The middle row is the one that matters here: the four tests this branch regressed are `@slow`, so
until this run they had never been executed by CI at all. They now pass on Linux as well as
locally. The visual pass is green too, which confirms the 1-px diff below is a macOS artefact and
nothing else.

Run locally on macOS at this SHA:

| Gate | Result |
|---|---|
| `npm run typecheck` | pass — 479 errors against a recorded baseline of 479, **no new type errors** |
| `npm run build` | pass — 1003 modules, 16.2 s |
| `npm run test` (unit + build passes) | pass — **7601 tests**, 404 files, 12 skipped, 1 todo |
| `cargo test --test paper_examples` | pass |
| `npm run test:e2e` (whole suite, not just `@smoke`) | **925 passed, 4 failed, 5 skipped** in 53.4 min — every remaining failure classified below, none of them this branch |

### The regression that held this PR in draft, and its fix

`ef79f8c2` closes it. For the record, because the shape is worth knowing:

F6 moved both standing notices onto the shared `RebarNotice` shell and left the state colour on a
wrapper it added outside them — `.provisional-band` and `.torsion-band` carried the background and
the bottom rule, while `data-testid` travelled inward with the notice.

```svelte
<div class="provisional-band">        <!-- background: var(--st-provisional-bg) -->
  <RebarNotice testid="rebar-provisional-banner" … />   <!-- .notice declared no background -->
</div>
```

**Nothing looked wrong, and that was the point.** `.notice` has `margin: 0` and fills the wrapper,
so the violet and the amber were painted over exactly the same pixels. What broke was the guard:
four tests read `getComputedStyle` of the element carrying the testid and found a transparent
`<p>` — expected `rgba(160, 102, 211, 0.16)`, received `rgba(0, 0, 0, 0)`; expected the warn token
`rgb(217, 164, 65)`, received `rgb(244, 247, 250)`.

| | |
|---|---|
| `e2e/status-token-consumers.spec.ts:256` | the provisional banner, en / es / pt — 3 tests |
| `e2e/viewer-panel-tokens.spec.ts:343` | the torsion amber — 1 test |

Both spec files are byte-identical to H1's, and all four pass on `feat/pro-concrete-h1`. That is
how the fault was isolated: same test code, different result, so the branch had changed what they
guard. `RebarNotice` had even written the intent it failed to keep — *"`data-testid` for the notice
itself, so existing specs keep their locators"*. The locators survived; what they pointed at
stopped carrying the state.

The fix moves the two declarations onto `:global(.notice)`, the element that is measured, for the
same reason the `<strong>` rules beside them already needed `:global`. **CSS selectors only** — no
markup, no testid moved in the DOM, no accessible structure touched, and no snapshot or timeout
rewritten. Each notice keeps its own palette, which is what F6 wanted; the wrapper remains as the
scoping anchor, which is the only job it ever had.

All four now pass here, and the two spec files run 18/18 — the same count H1 gives.

### The four remaining failures, none of them this branch

**1–2. `e2e/basic-selection-permutations.spec.ts:145`, 2D and 3D** — **pre-existing.** Reproduced on
the base `feat/pro-steel-family` (`a32607e8`), where it fails identically. Basic-mode selection, out
of scope for both H1 and H2. The spec carries no tag, so no CI pass ever runs it — which is why a
red test on the base went unnoticed.

**3. `e2e/f3-member-list.spec.ts:230`** (`@slow`) — **flaky under full-suite load.** A 60 s test
timeout inside `page.evaluate(solve())`. The whole spec file passes in 43.6 s on its own, 15/15.
A solve deadline exceeded while the machine runs everything else; the failure is a timeout, not an
assertion. No timeout was raised to make it pass.

**4. `e2e/rc-design-visual.spec.ts:21`** (`@slow visual baselines`) — **environmental.** A 1-pixel
width difference, 696 px expected against 697 px received, measured from macOS against Linux-only
baselines under `e2e/__screenshots__/linux/`. `continue-on-error: true` in CI by approved decision
O7. Also fails on H1.

**Not a failure: `e2e/h1b-panel-navigation.spec.ts:223` on H1.** It carries `test.fail()` — the
deliberate marker on the wrapped-strip chevron, left there because `WorkflowStages.svelte` is
shared PRO chrome H1 does not edit unilaterally. F1 makes the strip one row, so this branch removes
both the defect and the marker.

## Limitations

- **The `@slow` and visual E2E passes do not run on a PR** without the `run-e2e` label. Adding the
  label is not enough on its own: the workflow triggers on `pull_request`
  (opened / synchronize / reopened), not on `labeled`, so it takes a subsequent push to take
  effect.
- **Visual baselines are Linux-only**; a macOS contributor sees sub-pixel diffs that mean nothing.
- **`getDisplayRatio` returns `99` when utilisation is not computable** — see H3 below. It reaches
  the design table as a number and the 3-D viewer as a colour.
- **`designAll()` targets 0.95 utilisation by construction** (`certificate.designTarget`), so a
  freshly designed structure always reports ~0.98 and enlarging the section does not lower it.
  Correct behaviour, undiscoverable from the UI.

## Pending for H3

Surfaced by manual QA and by the §6.3 reconstruction. None of it is introduced by this branch and
none of it was fixed here.

1. **The `99` sentinel.** `verification.svelte.ts:459` returns `99` when `worstUtilization` is not
   finite. It is rendered as a D/C in `ProDesignTab.svelte:138` and fed to
   `verificationStateColor` in `viewport3d/results-sync.ts:478`. A D/C of 99 is a catastrophic and
   false claim; the engine meant "not computable". Preferred fix: return `null`, which the
   signature already admits.
2. **Flexural capacity is inflated by symmetric reinforcement.** With the tension face fixed, the
   reported negative-moment `φMn` for a 30×50 with 5 Ø16 moves 164.40 → 240.41 → 399.58 kN·m as
   the *compression* face goes 2 Ø10 → 5 Ø16 → 8 Ø25. The first value matches the classical
   calculation; the others do not. Factor 1.47–1.77, always in the direction of more capacity, and
   the module's own design path disagrees with its provided-reinforcement verifier. Shear is
   unaffected.
3. **A concrete material's `fy` field carries f'c.** Supplying the literal 420 classifies the
   material as steel, `computeDemands()` still returns `{ok: true}`, zero member contexts are
   built, and every member reports `unavailable` — the same word used for a member with no
   reinforcement. No error, no warning.
4. **A saved `.ded` does not persist `includeSelfWeight`** and reopens with it ON, changing ΣRz by
   31 % on the reconstructed frame without saying so.
5. **The share URL drops `reinforcement`.** Geometry, sections and loads survive; the design does
   not, so a shared RC model opens with every member `unavailable`.
6. **Untagged E2E specs never run in CI** — `basic-selection-permutations` is one, and there are
   more. A repo-wide decision, not H2's.

## Standing constraints observed

No `WorkflowStages`/`ProRibbon`, no M1, no V1/golden/Landing/Basic/Education, no
solver/Rust/Cargo/WASM product code — the only Rust here is one test file. No fabricated VERIFIED
results, progress, cancellation, store-unsupported states, or a retroactive `ExportRecord`. No
hidden conflicts or limitations. No `Co-authored-by`. Line ceilings not raised.

Out of scope for both branches and owned elsewhere: the C/Z mirror in `cold-formed.ts`
(M1 / integration), the three shared-chrome one-liners (integration), the 1172 missing `pt` keys,
and the remaining 462 `--st-text-3` sites.


diegokingston:
## `main` merged in — and one gate left red on purpose

Two commits pushed onto this branch (`7c0bbb36`, `80d980fc`). H1 landed on main via #191, so this branch's base moved out from under it. The CI run in the description above predates them.

### The four mechanical conflicts

**`e2e-hooks.ts`, `history.svelte.ts`** — additive on both sides. This branch's `rebarSelection` and snapshot `exports`/`manualEdits`; main's `selectedNodeIds`, `nodeMarkerRadius`, `jointMeshCount`, `jointDesigns`. All kept.

**`pt.ts`** — three-way on keys, not a union. Measured before resolving: this branch adds **188 keys, none already in main**, and changes **2 values, neither touched by main** — so no judgement call existed. Result asserted: **5281 unique keys** (5093 + 188), **zero duplicates**. A union would have been wrong for the same reason it was in #191, where it would have resurrected landing keys main had deleted.

**`DocumentsSection.svelte`** — this branch's side, since it moved the three exporters into `document-exports.ts`.

### `ProPanel.svelte`: main's side, and the gate that now fails

Both branches carry a phone panel — this is not "main added mobile". They diverged on its design, and main's is the shipped one (pill-and-grid, plus `473ba048` fixing the `ReferenceError` that made the whole PRO phone shell throw on mount). Main's side kept on all four hunks.

```
GATE: ProPanel was decomposed > stays inside the 600-LOC ceiling
ProPanel.svelte is 1098 lines: expected 1098 to be less than 600
```

**F0's decomposition survived** — the other three ProPanel gates pass, including *"delegates the catalogue, the overlay and the report assembly"*. What pushes the file over is main's phone shell living inline in it.

Left red rather than lowering a ceiling that is doing its job. The remedy it asks for is to extract that shell the way F0 extracted the rest, which is surgery on shipped mobile UI and your call, not a merge resolution's. `e2e/pro-mobile-shell.spec.ts` is what would verify it.

### A defect the merge would otherwise have buried

`withExportLog` is synchronous — correct for the two exports whose `run` writes a blob, wrong for `exportDetailingXlsx`, whose `run` returns a promise because `exportToExcel` is async. It logged **success the instant the promise was created** and returned; by the time it rejected the `catch` was gone. A failed spreadsheet export was written into the emission list as a spreadsheet that exists — the one thing that list must never say, and its own header claims *"recording is not swallowing"*.

main had fixed this from the other side (`exportXlsx` made async with an `await`), and **that fix did not survive moving the exporters into `document-exports.ts`**. Repaired in the wrapper so the next async exporter inherits it. Two tests, both directions verified — the thirteen pre-existing ones stayed green with it broken, which is how it survived.

### Gates after the merge

`typecheck` **473 = main's baseline**, no new · unit suite **8675 pass, 2 fail** (470 files): the ProPanel gate above, and `shed-bracing` timing out at 5 s under load — pre-existing on main, untouched here, fixed in #186.

#194 held these commits briefly and is closed; there was no reason for it to be a separate branch.

PR #183, opened 2026-09-03
PRO — 1/4: fixes, model integrity, and analysis features brought to the surface

Batuis:
**Part 1 of 4.** This is the first of four stacked PRs that take PRO from a tool that handles
exercises to one that holds up on a real project. The split is by size, not by taste: see
*How this is divided* below.

Draft while the blocks below are still landing.

## Why this programme

Two things decide whether PRO is usable on real work, and neither is the solver — the engine
already covers buckling, P-Delta, modal, spectral, pushover, harmonic, time history,
corotational, arc-length, material and fibre non-linearity, contact, staged construction, creep
and shrinkage, prestress, soil-structure interaction, influence lines and moving loads.

1. **A 500-member model is drawn one member at a time.** There is no way to repeat, mirror or
   rotate what you have already built, so model size is capped by patience rather than by the
   engine.
2. **A real project produces hundreds of load combinations**, and there is no way to say which
   of them a design run should consider, or to name an envelope and reuse it.

Everything in the four parts serves one of those two, plus the correctness work that has to
come first.

## How this is divided

The rule is **10 000 net new lines per PR**. Past that a PR stops being reviewable, so it
splits. Four parts, each comfortably under the ceiling:

| Part | Contents | Est. new lines |
|---|---|---|
| **1 — this PR** | Fixes already landed here, plus design-check correctness, model integrity checks, and analysis capabilities the engine has but the UI never exposed | ~8 000 |
| **2** | Geometry generation and editing: repeat, mirror, rotate, split, merge, intersect, move-with-connections, named groups | ~8 000 |
| **3** | Results at scale: active case list, named envelopes, summary and envelope views of every table, deflection relative to the chord, result annotation on the view | ~5 000 |
| **4** | Section optimisation — search for the lightest section that verifies and push it back into the model — and the view: saved views, multiple windows, granular labels | ~5 300 |

Parts 3 and 4 may merge into one if they come in under the ceiling. They are listed separately
so the ceiling decides, not the calendar.

## What is in this PR

**Already landed**

| | |
|---|---|
| `82d4eb37` `8371d3da` `4728cd41` | The bolted joint is drawn as the joint it is — oriented, perforated, inspectable — with a reachable viewer mode and no sphere over the plate. The contract is asserted by part rather than by a mesh total, so a change that moves geometry between parts cannot pass unnoticed. |
| `7dd48ddf` `4b402176` | The application's vertical axis is Z, and four places still said Y. The spreadsheet template now says Z too, which is where the app already had it. |
| `23a9c86e` `461ccfa0` | The manual QA inventory for the integration, and the full-suite result it was built on. |
| `5c3dca97` | **The doubly reinforced section must balance before it reports a capacity.** `computeFlexureCapacity` searched for the neutral axis by a fixed-point substitution that, with symmetric reinforcement, has no fixed point: it oscillated, ran out its twenty passes, and returned a state where Cc + Cs exceeded T by 47 %. φMn came back at 238 kN·m for a section carrying about 164 — always toward more capacity, and worse the more steel was placed. Replaced by bisection on a residual that is monotone in c, so the root is unique and always found, with equilibrium as a postcondition: if the balance does not close the member reports as not verifiable rather than returning a number that solves nothing. The flagship-frame digest is re-recorded with member-by-member evidence — across 408 members no outcome and no limiting constraint changed, 65 utilisations moved, worst 0.9970 against a gate of 1.0. |

**Still to land here**

- **Design-check correctness.** Four defects remain from the manual QA of the concrete flow; the
  fifth and only non-cosmetic one landed above. The rest: a `99` sentinel rendered as if it were a D/C ratio, a material field that
  silently classifies concrete as steel and leaves every member `unavailable` with no error, a
  project file that does not persist the self-weight switch and reopens 31 % heavier, and a share
  link that drops the reinforcement.
- **Model integrity checks.** Duplicate and orphan nodes, duplicate and zero-length members,
  overlapping collinear members, duplicate and warped plates, member/plate connectivity — found,
  listed, selected and removable in one step. Plus a **statics check table**: applied loads,
  reactions and their difference in all six components, per load case. Cheap to compute, and it
  is the fastest way to know a model still closes after an edit.
- **Analysis capabilities brought to the surface.** The engine supports these today and nothing in
  the UI reaches them: the seismic mass model assembled from chosen gravity cases with a factor
  each, the time-history definition (harmonic, tabulated, external file, or synthesised from a
  spectrum, with per-location arrival times), modal and Rayleigh damping, and a time-step slider
  over the response.

## QA

The manual QA already in this branch continues **in parallel** with the blocks above rather than
after them; each block closes with its own pass. The five design-check defects listed above are
the output of the last round, which is why acting on them is part of this PR rather than a
separate one.

## Gates

Every block closes on the full set before the next one opens: `npm run typecheck` against the
recorded baseline, the full unit suite, the production build, and the Playwright suite — not only
`@smoke`. No snapshot is rewritten and no timeout raised to make a failure pass; a failure is
either fixed, or reported and classified.

CI results will be recorded here as blocks land.


Batuis:
**Playwright, full suite, at `7d4f5e9d`:** 1332 tests. 1317 passed, 8 failed, 1 flaky, 5 skipped, 1 did not run.

Each failure was classified by re-running it on `main` (`3153eb84`) and, alone, on this branch:

| Test | On `main` | On this branch, alone | Cause |
|---|---|---|---|
| `pro-project-files` B: the file says PRO | — | passes | **This branch.** The test pins the saved file's top-level keys; `5ed91762` adds `includeSelfWeight`. List updated in `5e5d0c76`. |
| `ded-roundtrip`: the 7-storey project survives the file | — | passes | **This branch**, same key list. Updated in `5e5d0c76`. |
| `basic-demos`: drawing a beam | passes | passes | Load: the full run shared the machine with a unit-suite run. |
| `tab-reactivation`: 7-storey building | passes | passes | Load: a 6 s panel budget read 7.5 s under the same contention. |
| `i18n-languages`: the PRO workflow in en / es / pt | **fails** | — | Pre-existing: the project tab's Save button reads "Save", the test expects "Save Tab". |
| `pro-project-files` A: opens from PRO | **fails** | — | Pre-existing: the Open button has no `title`. |
| `rc-design-visual`: overlay legend | **fails** | — | Pre-existing, and non-blocking by its own declaration. |

No failure on this branch remains that does not also fail on `main`.

PR #188, opened 2026-09-09
feat(landing): a shorter deck, and a pricing section that says what will cost

Batuis:
The landing argued its case over **fourteen sections and 21,700 characters** of Spanish. It is now **eight and 4,700 — 78% less text** — and it says more, because what it says is what the product is and what it costs, rather than a defence of each capability in turn.

```
hero · Basic · PRO Analysis · PRO Design · AI · Education · Pricing · blog
```

## What changed

**PRO is two sections, not one.** The engine is free forever and the code-design modules probably are not. A single section covering both had to hedge to do it.

**Pricing carries the table from the one-page business model** — module, who it is for, price — and sits at the **end**.

It opened the page for one draft, under that document's own headline: *"what can be copied is free, what cannot is not."* That sentence is true and it belongs in an internal document. It explains the pricing from the company's side, and a visitor who meets it first reads a page thinking about revenue before it has said what the thing does. The deck now shows the product, then says what it costs, and frames it as an open project that invites participation.

**Deleted:** ten components and 214 translation keys across three languages.

## Contact, in both places

The corner button used to open WhatsApp — one channel presented as if it were *the* way to reach us. It now opens a small panel **on the page** listing all five (WhatsApp, Instagram, X, LinkedIn, Discord), each with its mark and, beside it, **the languages that account actually answers in**.

The language codes are the part that is not decoration. Someone writing in Portuguese to an account kept in Spanish gets silence and reads it as being ignored; `EN/ES` beside WhatsApp and `ES` beside Instagram costs a line and sets the expectation before the message is sent. On the page rather than at a link hub, because a visitor who presses a contact button has not asked to leave the site.

**The editor's header now offers the same five**, between Stabileo AI and Settings. That corner holds the controls that act on the application rather than on the document, which is what a contact menu is — and beside the assistant specifically, because the two answer the same impulse, one automatically and one with a person at the other end. Until now the way to ask a question vanished the moment someone stopped reading and started modelling.

The channels, the marks and the language codes are one description in `lib/contact/channels`, read by the landing's panel, the landing's footer and the editor's menu. They were about to be three copies of a list of handles.

### The phone, which this broke first

At 390px the header was already exactly as wide as the window: logo, mode selector and actions came to 387 of 390 available pixels. The new button pushed Settings off the right edge — still in the DOM, still focusable, unreachable by touch. Below 460px the wordmark beside the triangle goes, below 410px the word "AI" beside its own triangle goes, and the row's spacing tightens; both are labels on marks that already say it, and every button keeps its title and aria-label.

Still overflowing at 320px, as it was **before** this change (344px of content then, 349px now). Left alone rather than half-fixed: the width left is in the mode selector, and narrowing it truncates "Educativo (Beta)".

## The first screen

The nav is sticky, so it takes its own height in layout and then stays over the page; the hero added its full section padding underneath. The two stacked — the title opened 148px down and rose to 90px the moment anyone pressed the brand. The page shipped with a gap that the logo removed, which is a strange thing for a logo to do.

The hero's top padding now subtracts the header's height, and the hero carries a scroll margin of the same height so the brand lands where the page already was. That shortened the hero, and the section under it is paper: the first screen ended in a white strip. The hero now spans at least the screen minus the header, so the seam falls exactly where scrolling starts — `svh`, not `dvh`, so a phone's retracting address bar does not re-lay-out the hero mid-scroll.

## The languages

Written in all three, not only Spanish. `t()` falls back to English and then to the key itself, so Spanish-only copy would have rendered `landing.proCalcH` on the English and Portuguese landings — breaking exactly the indexable pages the SEO work just built.

**Spanish is the authored copy. English and Portuguese are mine and unreviewed.**

One deliberate wording change to flag: the Spanish H1 is now *"Cálculo estructural…"* rather than *"Análisis estructural…"*. It is the local phrase and the vocabulary the business-model document uses.

## Two gates moved, deliberately

`landing-i18n-parity` required more than 300 public keys, and a list of runtime-composed prefixes belonging to the capability matrix and the status ladder. Both encoded the old size. The threshold is 100 and the prefix list is `basicPt`. The shrinkage is the point of the change rather than an accident, and the assertions still catch a collapse.

The three contact strings moved from the `landing.*` namespace to `contact.*` now that two surfaces use them. That took them out of the landing's parity scan, so the contact test asserts directly that all three languages carry all three strings — and that none of them is still the English one.

## One thing I broke and put back

The hero figure's caption was rewritten along with everything else, and it described a figure I had not looked at: it is **a unit load travelling across a six-panel Pratt truss**, not "a nine-member truss under load". The original copy was exact about the load case, the colour scale and the normalisation. Restored verbatim in all three languages, and caught by the test that reads the figure's accessible description.

## Tests

Twenty tests asserted copy that no longer exists and went with it. What replaces them guards something harder than layout — this page states **prices**:

- the table rows match the business model exactly, and every priced cell carries its footnote
- the engine is promised free, and the design modules are free *while they are being built*, not forever
- the universities column of the table matches what the section promises, and the AI's
  cell is blank rather than priced, with the sentence that explains the blank present
  (it asserts both sentences are there; it does **not** assert that the free-for-universities
  sentence is qualified — see the review thread on `es.ts`)
- the AI section shows no screenshot of something that does not exist
- the five social links are off-site, `noreferrer`, and fetched from nobody
- each of the three languages renders all eight sections with no raw keys

Two assertions were folded into the hero's existing case rather than added as new ones — this suite runs every spec through a single browser and has wedged before near the end of a long run. The second of them runs at a 900px-tall window on purpose: at Playwright's default 720 the hero is already taller than the fold, so the check would pass against the exact defect it is there for.

## Base

Merged `origin/main` twice rather than rebased: the branch is twelve commits and the history
around it is full of merge commits, so a rebase would have rewritten twelve signatures to
avoid two merges.

- **First merge, 220 commits** — no conflicts, and it touched nothing under `components/landing`.
- **Second merge, 83 commits, during review** — one conflict, `e2e/landing.spec.ts`, three hunks.
  Main replaced the landing's native `<select>` with a drawn listbox and moved every language
  switch to `pickLanguage` / `data-value`; this branch rewrote the H1 in Spanish and Portuguese.
  Each hunk paired one change with the other's stale half, so neither side was right whole.
  All three now take main's interaction with this branch's copy. `@landing` does not run in CI,
  so taking either side whole would have stayed green there.

## Deploy

The prerender verification in `deploy-gh-pages.yml` grepped one hero sentence per language in
`dist/`, and this branch rewrites `landing.heroP` in all three. That workflow runs only on push
to `main`, so nothing on this PR could have shown it: the first deploy after merging would have
stopped at that step.

It now checks `<html lang="…">`, which `scripts/prerender.ts` writes from the captured language.
The prose stays asserted per language by `prerender.spec.ts`, which runs in `@smoke` on every PR.

## Accessibility

The editor's contact menu declared `role="menu"` / `menuitem` without the keyboard contract those
roles promise — arrow keys, Home, End, typeahead. It is a disclosure over five ordinary links and
now says so: `aria-expanded`, `aria-controls`, and a labelled `role="group"`, the shape the
landing's corner panel already used. Escape returns focus to the button in both.

## Gates

`test:unit` · `typecheck` at baseline 472 · `check:gate` · `build` with prerender · `@smoke` + `@landing` e2e.

Re-run after the second merge: `typecheck` 472 against a baseline of 472, and `landing.spec.ts` +
`contact-menu.spec.ts` + `prerender.spec.ts` at 54 passed.


web/src/components/landing/LandingPricing.svelte
```diff
@@ -0,0 +1,113 @@
...
+  const rows = [
+    { mod: 'modBasic',     firms: FREE, unis: FREE },
+    { mod: 'modProCalc',   firms: FREE, unis: FREE },
+    /*
+     * Free while it is being built, for everyone — not "paid, at low prices".
+     * The note the asterisk points at carries the rest: early users keep it,
+     * public universities always do, and what comes later is cheap. No date
+     * is published, deliberately.
+     */
+    { mod: 'modProDesign', firms: { k: 'priceFreeDev', tone: 'free', note: 1 }, unis: FREE },
```
diegokingston, line 56: **Decision needed — are the design modules free for *all* universities, or only public ones?**

The page currently says both:

- **This row:** the universities cell is `FREE` → "Gratis", with **no footnote marker**.
- **The column header** is "Universidades" (all), and `pricingP` ends *"En cualquier caso, el programa es gratuito para las universidades"*.
- **The e2e test** asserts it: *"public and private were separate rows and everything is free for both."*
- **But note `*`** (`pricingFreeNowNote`, `es.ts:10`) says the design modules will be free *"para las universidades **públicas**"*, and that note hangs off the **firms** cell only.

A private university reading this row sees "Gratis" with no asterisk, and the only sentence that excludes it points at a different cell. The same happens in `en` and `pt`.

Whichever is true, the table, the lead paragraph, the note and the test should say the same thing. If the intent is public-only, this cell probably needs its own marker.

web/src/lib/i18n/locales/es.ts
```diff
@@ -2,19 +2,112 @@ const es: Record<string, string> = {
...
+  'landing.shotEduAlt': 'Un ejercicio en el modo educativo: el pórtico a la izquierda y, a la derecha, los pasos con los campos de reacciones que el alumno completa.',
+  'landing.shotEduTitle': 'Un ejercicio de Estática',
+  'landing.shotAiTitle': 'Propone el cambio; vos lo aplicás',
+  'landing.shotAiAlt': 'Panel de Stabileo IA: un pedido en lenguaje natural y el cambio que la IA propone antes de aplicarlo.',
+  'landing.ebPricing': 'PRECIOS',
+  'landing.pricingH': 'Un desarrollo abierto con varias aristas',
+  'landing.pricingP': 'Stabileo es open source y se construye con la comunidad: el código está a la vista, se puede auditar y cualquiera puede participar. Queremos que todas las herramientas de cálculo tradicionales sean de acceso libre, y que las de diseño específico y Stabileo IA eventualmente ayuden a financiar el proyecto. En cualquier caso, el programa es gratuito para las universidades.',
```
diegokingston, line 22: **Decision needed — the lead promises universities "the program" free, and Stabileo AI isn't.**

*"En cualquier caso, el programa es gratuito para las universidades."* carries no exception, while the AI row's university cell is blank and note `**` says the AI is a paid feature (universities can plug in their own model API).

The PR description says a test ensures that *"the promise to universities never appears without its AI exception"*. What the test actually checks is that both sentences are **present** in the section (`gratuito para las universidades` and `conectar la API del modelo de lenguaje`). It does not check that this sentence is qualified.

Two options:

- **Qualify the sentence itself**, e.g. *"…gratuito para las universidades, salvo Stabileo IA"*.
- **Narrow it** to what the table actually offers.

The same applies to `en` and `pt`.

web/src/lib/i18n/locales/es.ts
```diff
@@ -2,19 +2,112 @@ const es: Record<string, string> = {
...
+  'landing.basicH': 'La mejor calculadora de bolsillo open source de la ingeniería civil.',
+  'landing.basicP': 'Modelás, cargás, resolvés y leés resultados en 2D y 3D, en el navegador.',
+  'landing.basicPt1': 'Resolvé estructuras de barras en 2D y en 3D.',
+  'landing.basicPt2': 'Analizá esfuerzos característicos y tensiones.',
+  'landing.basicPt3': 'Probá las funciones de análisis avanzado: cinemático, tensiones en una sección y varias más.',
+  'landing.basicPt4': 'Cargá modelos desde plantillas de Excel, guardalos o compartilos con un link.',
+  'landing.ebProCalc': 'PRO · CÁLCULO',
+  'landing.badgeDevFree': 'EN DESARROLLO · GRATIS',
+  'landing.proCalcH': 'Elementos finitos y modelos complejos.',
+  'landing.proCalcP': 'Losas y tabiques, construcción por etapas, no lineal, dinámica. Al nivel que esperás de un paquete profesional, y ya funcionando.',
```
diegokingston, line 79: **Decision needed — "in development", or "already running at the level of a professional package"?**

This paragraph says *"Al nivel que esperás de un paquete profesional, y ya funcionando."* Right above it, the badge says **EN DESARROLLO · GRATIS**, and the caption under the screenshots says *"Está en desarrollo"*. The section contradicts itself within three lines.

It also lists "construcción por etapas" at a professional level. Staged construction silently dropped every slab and wall until #160, which merged three weeks ago. The claim may well hold now, but it is exactly the kind of sentence the removed status section (`LandingStatus`) was there to keep honest.

Two options:

- **Align the paragraph with the badge.**
- **Drop the badge**, if "already running" is the intended position.

The same wording appears in `en` and `pt`.

web/src/lib/i18n/locales/es.ts
```diff
@@ -2,19 +2,112 @@ const es: Record<string, string> = {
...
+  'landing.ebProDesign': 'PRO · DISEÑO',
+  'landing.badgeDevOpen': 'EN DESARROLLO · ACCESO LIBRE',
+  'landing.proDesignH': 'Verificación normativa y diseño de elementos.',
+  'landing.proDesignP': 'Después de calcular y obtener los resultados, diseñá los elementos que quieras según el reglamento que corresponda. Hoy está adaptado a la normativa argentina (CIRSOC) y preparado para incorporar los reglamentos de cualquier otro país.',
```
diegokingston, line 83: **Decision needed — does "adaptado a la normativa argentina (CIRSOC)" cover steel?**

As written, this reads as covering both CIRSOC 201 (concrete) and CIRSOC 301 (steel). On `main`, both of these say otherwise for steel:

- **The code:** `web/src/lib/engine/steel/steel-status.ts` still classifies steel design results as `EXPERIMENTAL`.
- **The blog:** the published post on the same domain, `determinism-boundary.ts:111`, says: *"el de acero vive todavía sólo en la capa de frontend y no tiene batería de tests dedicada, y se dice así para no atribuirle una madurez que no tiene."*

So the landing would claim a maturity that the blog explicitly disclaims. I'd suggest distinguishing the two, e.g. *"hormigón (CIRSOC 201) implementado y testeado; acero (CIRSOC 301) experimental"*. The same applies to `en` and `pt`.

web/src/lib/contact/channels.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * What language each channel is actually written in.
+ *
+ * Not translated, and not derived from the reader's locale: these describe the
+ * ACCOUNT, so they read the same in every language. A Portuguese reader learns
+ * that Instagram is kept in Spanish, which is the point — the alternative is
+ * following an account they cannot read, or writing a message that gets an
+ * answer they cannot read.
+ */
+export const CHANNEL_LANGS: Record<string, string> = {
```
diegokingston, line 69: **Please confirm — are these the languages each account is actually run in?**

```ts
whatsapp: 'EN/ES', instagram: 'ES', x: 'EN', linkedin: 'ES', discord: 'EN/ES'
```

These codes are the point of the panel: they set the reader's expectation before they write, which is exactly the argument the PR description makes. A wrong code creates the very expectation the feature is meant to prevent. For example, if LinkedIn also posts in English, `ES` tells an English-speaking reader not to bother.

Worth checking against how each account is really managed, particularly `x: 'EN'` and `linkedin: 'ES'`.

diegokingston (commented):
Review pass. I pushed the mechanical fixes to this branch. The copy and pricing points need a product decision, so they're left as inline comments on the lines they concern rather than changed.

### Pushed

1. **Merged `main`** (83 commits) and resolved `web/e2e/landing.spec.ts`.
   - Each of the three hunks paired main's new language control (a drawn listbox driven through `pickLanguage` / `data-value`, from 2d35713d) with the old H1 copy, against this branch's new copy with the removed `<select>`. Neither side was right on its own.
   - All three now use main's interaction with this branch's copy.
   - `@landing` doesn't run in CI, so a resolution that took either side whole would have stayed green there.
2. **`deploy-gh-pages.yml` would have failed on the first deploy after merge.**
   - The prerender verification grepped one hero sentence per language in `dist/`, and this branch changes `landing.heroP` in all three. That workflow only runs on push to `main`, so nothing on this PR could have shown it.
   - It now checks `<html lang="…">`, which `scripts/prerender.ts` writes from the captured language. The prose stays asserted per language by `prerender.spec.ts` in `@smoke`.
   - Those prose checks were added in #162, alongside the same strings in `prerender.spec.ts`. Keeping them in two places is what let one copy go stale. The workflow already records the same failure mode once: a literal check went stale after #172 and blocked deploys until someone found it.
3. **Contact menus.**
   - The header menu declared `role="menu"` / `menuitem` without the keyboard behaviour those roles promise. It's now a disclosure (`aria-expanded`, `aria-controls`, a labelled `role="group"`), matching the landing's corner panel.
   - Escape now returns focus to the button in both.

### Verified locally

- `npm run typecheck`: 472 errors, the same as the existing baseline. None are new.
- Unit (`vitest --project unit` over `src/lib/contact`, the landing tests and the i18n tests): 17 files, 233 passed.
- e2e `contact-menu.spec.ts`, `prerender.spec.ts`, `landing.spec.ts`, including the `@landing` suite CI skips: 54 passed.
- Deploy: the workflow's *Verify the public pages were prerendered* step, extracted verbatim and run against a fresh build, passes. The three old prose checks on the same build fail 3 of 3.

### Decisions (inline comments below)

- Design modules: free for all universities, or only public ones? (`LandingPricing.svelte`)
- The lead says the program is free for universities, but Stabileo AI is paid. (`es.ts`)
- PRO · Cálculo: "in development", or "already running at a professional level"? (`es.ts`)
- CIRSOC scope: concrete (201) is implemented, steel (301) is still experimental. (`es.ts`)
- Channel language codes: do they match how each account is run? (`channels.ts`)

### Also worth a look (not blocking)

- **The business model isn't in the repo.** The pricing test compares the table against rows typed into the test, so "matches the business model exactly" currently means it matches itself. Committing at least the table under `docs/` would give the test a real source and let reviewers check it.
- **`basicH` makes an unverifiable superlative claim:** *"La mejor calculadora de bolsillo open source de la ingeniería civil"*. The blog posts go out of their way to name competitors fairly.
- **"Pago por token" prices something that can't be bought yet.** Stabileo AI's client on `main` still defaults to `localhost:3001`. The section does carry an EN DESARROLLO badge.
- **The *Base* section of the PR description is out of date.** It says "220 commits, no conflicts", but `main` has since moved 83 commits and the conflict above appeared.

web/src/lib/contact/channels.ts
Batuis: Confirmed by the account owner, and they were not a guess on my part: the five codes were specified with these exact values when the panel was asked for — WhatsApp EN/ES, Instagram ES, X EN, LinkedIn ES, Discord EN/ES. They are the languages each account actually posts in.

No change here.

web/src/lib/i18n/locales/es.ts
Batuis: Leaving this as it is, deliberately.

Splitting the sentence into "201 implemented and tested / 301 experimental" would say the concrete half is finished, and it is not. The position is that PRO Design as a whole is still being built — which is what the section already states, in the badge next to the heading: **EN DESARROLLO · ACCESO LIBRE** (`badgeDevOpen`, and the same in `en` and `pt`). A reader does not meet the CIRSOC sentence on its own; they meet it under that badge.

The evidence you raise is correct — `steel-status.ts` classifies metallic design as `EXPERIMENTAL`, and the blog post says so in as many words. That is the finer-grained statement, and the blog is where it belongs. Re-stating the split on the landing would trade one over-claim for another.

web/src/lib/i18n/locales/es.ts
Batuis: Leaving the paragraph and the badge as they are.

"In development" and "already running" are not in contradiction here: the badge sets the maturity, the sentence says the capability exists today, and the caption repeats the maturity under the screenshots. That is the intended reading, and the staged-construction fix you cite is an argument for keeping the badge — which is what it does.

No change.

web/src/lib/i18n/locales/es.ts
Batuis: Two halves to this one.

**The sentence stays unqualified for now.** Adding "salvo Stabileo IA" is a commercial decision rather than a wording one, and it is not being taken on this PR. The note under the table already carries what a university can do about the agent — connect its own model API.

**Your reading of the test is right, and the PR description was wrong.** It claimed the test ensures the promise never appears without its AI exception. What the test does is assert that both sentences are present in the section, which is not the same thing. The description has been corrected to say exactly that, and to point here.

web/src/components/landing/LandingPricing.svelte
Batuis: Acknowledged and parked, not fixed — and it is worth being precise about what is left.

The decision taken is that the page does not add a qualification: no asterisk on the universities cell, and no narrowing of the lead. What that leaves is not a missing clarification but two statements that disagree — the column header, the table and `pricingP` say "universidades", and note `*` says "universidades **públicas**", in all three languages.

Under "do not qualify", the word that is out of place is `públicas` in the note. That is a one-line change per language, and it is not being made here because it widens a commercial promise to private universities, which is a call for the founders rather than for this PR. Tracking it so it does not get lost; it should be settled before the next pricing change rather than left to the reader to resolve.

PR #194, opened 2026-09-10
H2 to main: main merged in, with one gate deliberately left red

diegokingston:
**#170 (H2) with `main` merged in.** Raised against `feat/pro-concrete-h1`, which has since landed on main, so this resolves the five conflicts and the one consequence they expose.

Open it alongside #170 — the 66 commits of H2 are reviewed there. What is new here is the merge.

## The four mechanical conflicts

**`e2e-hooks.ts`, `history.svelte.ts`** — additive on both sides. H2's `rebarSelection` hook and its snapshot `exports`/`manualEdits`; main's `selectedNodeIds`, `nodeMarkerRadius`, `jointMeshCount` and `jointDesigns`. All kept.

**`pt.ts`** — three-way on keys, the method #191 established, because a union is wrong here for the same reason it was there. **Measured before resolving:** H2 adds 188 keys, **none** already in main, and changes 2 values, **neither** touched by main. No judgement call existed to be made. Result asserted rather than eyeballed: **5281 unique keys** (5093 + 188), **zero duplicates**.

**`DocumentsSection.svelte`** — H2's side: it moved the three exporters out to `document-exports.ts` while main's side still had them inline. Main's only edit to that file was making `exportXlsx` async — which is the second commit here.

## `ProPanel.svelte` — main's side, and the gate it breaks

Better than it first looks: **both branches carry a phone panel.** This is not "main added mobile, H2 removed it" — they diverged on its design, and main's is the newer one (pill-and-grid, plus the `473ba048` fix for the `ReferenceError` that made the whole PRO phone shell throw on mount). Main's is what ships, so main's side is kept on all four hunks.

**One gate is left failing, deliberately:**

```
GATE: ProPanel was decomposed > stays inside the 600-LOC ceiling
ProPanel.svelte is 1098 lines: expected 1098 to be less than 600
```

H2's decomposition survived — the other three ProPanel gates pass, including *"delegates the catalogue, the overlay and the report assembly"*. What pushes the file over the ceiling is main's phone shell living inline in it.

The remedy the gate is asking for is to extract that shell the way F0 extracted the rest. That is surgery on shipped mobile UI and a decision for its author, not something a merge resolution should decide. `e2e/pro-mobile-shell.spec.ts` is what would verify it. **Left red rather than papered over, and rather than lowering a ceiling that is doing its job.**

## A defect this merge would otherwise have buried

`withExportLog` is synchronous: `run()`, log, return. Correct for the two exports whose `run` writes a blob synchronously. `exportDetailingXlsx` hands it one that returns a **promise** — `exportToExcel` is async — so the wrapper recorded **success the instant the promise was created** and returned. By the time it rejected, the `catch` was gone: the rejection went unhandled and the emission list said a spreadsheet had been written.

That is the one thing that list must never say, and its own header claims otherwise: *"the error is re-thrown to the caller so the panel can show it; recording is not swallowing."* True of two exports out of three.

`main` had reached the same fault from the other side, making `exportXlsx` async with an `await`. **That fix did not survive H2 moving the exporters into `document-exports.ts`** — exactly the kind of loss a 183-file merge makes invisible. Repaired in the wrapper rather than at the call site, so the next async exporter inherits it.

Two tests, both directions verified: forcing the async branch off turns exactly those two red and leaves the other thirteen green. **The thirteen pre-existing tests stayed green with this broken**, which is how it survived.

## Gates

`typecheck` **473 = main's baseline**, no new type errors.

**Unit suite: 8675 pass, 2 fail** (470 files).

| Failure | Status |
|---|---|
| `rc-design-gates` — ProPanel 600-LOC ceiling | **The one above.** Deliberate; needs the phone shell extracted. |
| `shed-bracing` — `Test timed out in 5000ms` | Pre-existing on main, untouched by H2, fixed in **#186**. |

Nothing else in the suite moved.

## Order

**#192** stacks on H2 and carries two documentation fixes from reviewing #170. **#190** is independent and already targets main.

https://claude.ai/code/session_01L67REmkuj14xQGpVujp5z9


diegokingston:
Superseded: these two commits are now on `feat/pro-concrete-h2` itself, so **#170** carries them and stays the single PR for H2.

This branch existed for no good reason. #191 needed one because `feat/pro-steel-family` was checked out in another worktree and could not be pushed to; I reused the pattern here without rechecking that the reason still applied. It did not — `feat/pro-concrete-h2` was free, and the update was a fast-forward.

Splitting them cost something real: #170 carries the audit findings, the eight phases, the regression diagnosis, the classified e2e failures and the green `run-e2e` CI run, and this PR's description covered five merge conflicts. Keeping both would have meant neither was complete on its own.

Everything described here now applies to #170:

- the four mechanical conflicts (`e2e-hooks`, `history`, `pt.ts` by three-way key merge, `DocumentsSection`)
- `ProPanel.svelte` resolved to main's side, and the 600-LOC gate left deliberately red because main's phone shell lives inline in it
- the `withExportLog` fix: a failed XLSX export was being logged as a spreadsheet that exists

PR #196, opened 2026-09-11
Section reinforcement design: the half of #189 that never got pushed

Batuis:
Fourteen commits that were written but never pushed, so #189 merged only the first half of this work — plus everything that came out of testing what it restored.

## What was live on `main` and is fixed here

**A spiral column got ACI's φ, not CIRSOC's.** `phiC = spiral ? 0.75 : 0.65`; CIRSOC 201-2005 §9.3.2.2 gives a spiral 0.70. Capacity read about 7 % high — the unconservative direction.

**The rectangular column sized off a straight line**, interpolating the interaction diagram instead of reading it. A nine-point sweep put the result between 71 % and 171 % of the steel actually required.

**A band of moments had no design at all.** On a 20 × 50, Mu = 215 kN·m did not verify and Mu = 220 did.

**The φ ramp was hardcoded for fy = 420.** `losas.ts` read `epsilonT >= 0.0021` literally, so any other grade transitioned at the wrong strain.

## The reinforced-concrete calculator

Renamed — it is not tied to one workbook — and the first control is a code selector defaulting to CIRSOC 201-2005. Later editions and the foreign codes are listed and disabled with a reason rather than hidden; `design-codes.ts` keeps "what codes exist" apart from "what this can apply", so no edition is ever run quietly under another's name.

One engine — polygon clipping plus strain compatibility — serves all eight sheets. It is first made to reproduce the closed forms it replaces, so later agreement with the workbook is evidence about the workbook and not about the clipper. All eight sheets agree within 2 %, plus eighteen published verification numbers and six characteristic points per sheet.

Found by sweeping design against verification rather than checking the eight published points and stopping:

- **A column at zero axial load was credited with the nose of its diagram.** `utilisation` special-cased pure bending by taking the greatest moment anywhere on the curve, which is the balance region. A 30 × 30 asked for 100 kN·m wanted 16.3 cm² at Pu = 0 and 24.5 at Pu = 2 kN — and the smaller number was the wrong one. Pure bending is where the curve crosses zero, which now reproduces the workbook's published 87.79 kN·m to 0.03 %.
- **Checking a beam never looked at the steel.** `solveFlex` branched on mode for the column cases and not for the beams, so a check reported the ratio between a demand and a section sized to match it — necessarily 1.000. Two cm² against 150 kN·m came back "verifica".
- **A beam that needed more steel than one layer holds was refused.** Bars now stack, `d` follows the group centroid, and the two are solved together because each depends on the other.
- **`AsMinCm2` was evaluated on a width a circular section does not have**, so the panel printed "NaN cm²".

The memo now reads in the reader's language: it was Spanish whatever the locale said, because `solveFlex` pushed sentences instead of `{ key, params }`.

## Excel

The importer can describe any 2D or 3D model, with a drift guard that fails if the importer, the `.ded` file or the share link falls behind the app. A second guard sweeps every column of the template: change that cell and the import must change, or be refused. It found four columns — `mx`, `my`, `qzi`, `qzj` — that only the 3D load types read and that the template never demonstrated.

## Basic UI

Move returns to the ribbon with a panel that earns it: a drag can move the view or the model, and the second used to live inside the Node tool's create-mode, where the same gesture placed a node whenever the press missed one.

The ribbon and the right panel now agree — a tool command lights on what the panel is showing, not on where the pointer happens to be. A mouse click no longer leaves a focus ring behind while the keyboard keeps one. "Seleccionar barras" stopped selecting nodes, and node mode started being able to select them: both hit-tested at the snapped point rather than the cursor, as did double-click, which opened the bar editor for a node.

Nine camera buttons over a 3D model became two. Those controls were still on the old palette, and the backdrop that closes the new menu inherited a `:hover` rule that painted the entire application blue. A plan view swapped `camera.up` to Y — which reorients the camera and does nothing to the orbit axis, since `OrbitControls` fixes that in its constructor — so screen-up and orbit axis disagreed and the first drag rolled the model over.

The double-click editors are on the current tokens, draggable, and a member edit now goes through a store method: assigning onto the element object changed the data and told nothing, so stale results stayed on screen beside a member whose section had changed.

Project reads as a hierarchy, Import sits above Export, saving asks what and where, and the language menu opens over itself instead of wherever the platform put a native popup.

## Two things this branch broke and then fixed

**The language menu stopped being a `<select>`, and five specs still waited on one.** `selectOption` and `toHaveValue` do not resolve against a drawn list, and the failure arrives as a sixty-second timeout on a locator, which says nothing about the cause — the whole of the red `e2e` job was this. They press the button and pick by CODE now, through one helper: the options are language names written in their own language, so matching by text asserts a dictionary entry every time it means to assert a locale. `LocaleSelect` carries `data-value` for exactly what `toHaveValue` used to give for free.

**A test was failing at random and was not flaky.** `design-and-verify-agree`'s circular sweep takes 4954 ms against a 5000 ms default, so a loaded machine times it out — and vitest reports a timeout the way it reports a failed assertion, with the test's own message attached. It read as shared state in a module that reads none. The timeout is raised rather than the sweep thinned: the density is the point, since the defects it catches are a few kN·m wide.

## Checks

9078 unit tests and the e2e suites pass on top of current `main`. Typecheck holds its 473-error baseline. The solver is untouched.


diegokingston:
## Two fixes pushed, one held back, and a gap worth a decision

Picking up the round-2 work from `pr196-fixes-v1`. Its older commits are already here under different hashes (`87318b8a`, `4520191f`); only the three from Sept 14 were outstanding.

### Pushed (`20a21c3e`, `ee703f3e`)

`b5522480` — verification capacities use the steel's real stress, and the axial cap stays at the top.
`5d7fb89a` — the φ ramp lives in exactly one place. Conflicted on an import only (`ES_MPA`).

**276/276 across `codes/argentina`, typecheck 473 = baseline.**

### Held back: the biaxial load contour (`adec4346`)

It does not replay, and the reason is worth knowing: **this branch reworked the same function** while the fix was being written. The `Pu ≈ 0` branch of `utilisation` here finds the crossing and interpolates, with the workbook rationale in the comment (87.79 kN·m of pure flexure for 21.34 cm²); `adec4346` still has the earlier max-capacity-point approach, whose variables (`m`, `best`) do not exist in this scope. Taking its side would not compile AND would revert a validated correction; taking this side drops the contour.

So I ported it by hand instead: the named `Utilisation` type with `inside`, `centrallySymmetric` / `loadContour` / `utilisationBiaxial` verbatim, `passes()` on all four comparisons, keeping this branch's crossing and its i18n `msg()` calls. **Its own 13 tests pass.**

It is held back because it breaks an invariant of yours:

```
FCO: 24.6 22.0 19.5 17.1 20.8 21.3 40.5 24.4 …
                                    ^ three turns, two allowed
```

Probed with the real fixture (`Pu: 500, Mu: 100, Muy: 0`, 30×30): at **Pu = 600 the design lands at 40.5 cm² with ratio 0.696**, while its neighbours converge on 1.000. A ratio of 0.696 at the answer means the bisection rejected smaller areas whose ratio was already ≤ 1 — i.e. `inside` was false across a band.

**Likely cause.** `loadContour` only pushes a vertex where the curve crosses `Pu`; at angles where it never reaches it, nothing is pushed, and the only guard is `vertices.length < 3`. A contour missing a run of angles passes that guard and yields a polygon that does not enclose the origin, at which point the ray parity means nothing. Being area-dependent explains why `inside` only recovers at 40.5.

Note the failing demand is `Muy = 0` — axis-aligned, which `loadContour`'s own comment identifies as the case whose ray lands on vertices.

**Worth saying: `design-and-verify-agree.test.ts` does not exist on `pr196-fixes-v1`.** You added it here, after that branch forked, so the contour work has never met this invariant. This port is the first time the two ran together — the failure is as much a discovery as a regression.

I did not guess a fix. What an angle with no crossing MEANS — zero capacity there, skip it, or refuse the contour — is a numerical convention for a code-checking engine, and not mine to invent.

The port is on the stash `pr196-biaxial-port-wip-1789400522` (`c27541ef`), complete, with its test file, ready to resume once that is settled.

### The gap underneath it

**The genuinely skew path is the least-verified part of this engine.**

FCO's workbook agreement is `Pu: 500, Mu: 100, Muy: 0` → `Ast = 21.3515`. With `Muy = 0` the calculation degenerates to uniaxial: it is an axis-aligned demand on a biaxial sheet. The only test of real skew bending is

```js
expect(skew.AstCm2).toBeGreaterThan(axis.AstCm2);
```

which almost any implementation satisfies, correct or not — it cannot tell the ray approximation from the load contour.

So the sheet that exists FOR skew bending carries no published number, in an engine where everything else does. That is not a coincidence next to the item above: the path with no numerical agreement is the path with the open defect. **A published point with `Muy ≠ 0` would be worth more than either — without one, any contour implementation is only verified against itself.**

### Verified while reading, for the record

- `PHI_COMPRESSION_SPIRAL = 0.70`, with ACI's post-2008 0.75 named and refused. Correct.
- The clipper is pinned against closed forms at 12 decimals, so the 2 % workbook agreement really is evidence about the workbook. The layering is right.
- The template drift guard's exemption list is two named entries with the reason written. Exactly the right discipline.

Not reviewed: `CirsocFlexPanel.svelte` (1009 lines), `ToolbarProject.svelte` (559), the e2e specs.

PR #197, opened 2026-09-12
PRO interface, and Educational as a panel inside Basic

Batuis:
Targets #196, which carries the reinforced-concrete calculator and the Basic UI work this builds on.

## PRO

**Conditions were a stage and are not one.** Supports, constraints and loads sat beside Model and Analyse at the top level, which said they were a phase of the work. Nothing is produced by moving to them and nothing follows from leaving: they describe the structure, like its geometry and its materials. A reader building a frame crossed between two top-level stages to place a support and crossed back. They are one group inside Model now, right of Properties and left of Generators.

**Settings open in the right-hand panel**, like Basic's. PRO hung them off the header button in a dropdown — a second surface with its own scroll, close and focus trap, for content the panel already exists to show. Two ways of showing one thing is how the two modes drift. The gear is a toggle over the panel now — `aria-pressed`, not `aria-expanded`, because it owns no popup — and picking another destination in the ribbon leaves settings. Its four specs move with it: they keep the assertion worth having (the panel is measurably on screen with real content in it, which is the bug they were written for — a box with no `display: none` and its top below the fold) and drop the popover behaviour nothing implements any more.

**Project follows Basic's order.** One Save that asks what and where, rather than two buttons that each wrote a `.ded` on press with the difference explained nowhere. Then Import / Export, in that order, because a model arrives before it leaves — Import carries what Basic has plus PRO's DXF plan and IFC. Export calls the same functions the results table and the viewport call, so it is one route with two doors rather than two to keep in step. "Documento abierto" and "Guardado automático" stay and move below the actions: they are facts, not things you do, and they were opening the panel with two read-only cards above every button.

**The grid reaches ten kilometres and opens at one.** Fifty metres of floor ran out before a building did. The control is stops rather than a linear track, and `updateGrid` gained a line budget — ten kilometres at one-metre spacing is ten thousand divisions rebuilt whenever the working plane moves, for lines nobody can resolve at that zoom.

**D_agg and Margen were being asked of steel.** Both belong to detailing reinforcement in concrete, where §25.2 makes the clear spacing the larger of the bar, 25 mm and 4/3 of the aggregate. "Acero A36" carried an aggregate box — not a harmless blank but a question with no answer. They also looked the way they did because they were bare `<input>`s, drawn white-on-white in a dark panel.

**Sections list by name, with a detail worth opening.** A wide table of exponents is not how anybody recognises a section. The detail adds what a steel check is written in and which appeared nowhere in this application: the elastic moduli Wy and Wz, the radii of gyration, and the overall depth and width. They are derived from the properties the solver holds, so they cannot disagree with it, and are omitted where the geometry is unknown — a modulus from a guessed depth is worse than no modulus. Materials gained `G = E / 2(1+ν)` and the material FAMILY with the basis it was inferred on.

The dialogs behind "new section" and "new material" are untouched: they build composite and parametric sections Basic cannot express, which is the difference that should stay.

**Shell results stop being one colour per element.** The solver reports most components per element and a flat colour per element is the honest minimum — it is also a mosaic whose edges are mesh artefacts. Values are averaged at shared nodes and interpolated across the face, which is what a post-processor does with element results. The averaging is stated where it happens.

**Selection can take several kinds at once.** Right for a click to take one — the most specific thing under the cursor — and wrong for a marquee, where the reason to drag a box round a bay is to catch its members and the loads on them.

**Repeat selection**, which is what copy-and-paste was being used for. The clipboard pastes one copy at a fixed offset; six storeys are a count and a spacing. The part that is not copying is the members BETWEEN copies — repeating a floor plan gives you floors and no building — so `link` creates them, chained rather than fanned, because joining every copy to the ground would put three members through one space and analyse as three columns. It does not weld coincident nodes and says so.

## Educational

It sat at the top level beside Básico and PRO, claiming to be a different application. It is not: same model, same solver, same canvas, with a panel that turns an exercise into a sequence of questions — and the cost fell on the student, who had to leave the exercise to check anything the ordinary way.

Basic can now ask for the panel: a checkbox at the foot of Settings, off by default, remembered per browser. Switched on, a command appears at the right-hand end of the ribbon and opens `EducativePanel` where every other panel opens. The panel is mounted, not reimplemented — a second exercise surface is what a teacher and a student must not be able to disagree about.

First approach, as asked. What it does not yet do is constrain; that shape needs the accounts that do not exist yet, and nothing here forecloses it.

## Also

`docs/proposals/pro-modelling-tools.md` sets out what to build next for PRO's modelling tools and why, in order: merge coincident nodes first, because it is the repeat tool's stated limitation and two nodes in one place analyse as two nodes — a structure cut where it looks joined, with no visible symptom.

**A test was failing at random and was not flaky.** `design-and-verify-agree`'s circular sweep takes 4954 ms against a 5000 ms default, so a loaded machine timed it out and reported it as a broken invariant. Chased as shared state first: the module is pure and passed six times of six alone, which was the clue. The timeout is raised rather than the sweep thinned — the density is the point, since the defects it catches are a few kN·m wide.

## Checks

9090 unit tests pass. Typecheck holds its 473-error baseline. The solver is untouched.

---

# Second round

## The defect that mattered most

**Clicking a node worked or not depending on how the model got there.** The
modelling flow is: type the nodes with their coordinates in the panel, then
click those nodes to lay members and supports on them. The second half did not
work on a model built that way, and did work on one loaded from a file — the
same gesture, succeeding or failing for reasons invisible to the person making
it.

`InstancedMesh.raycast` tests the mesh's bounding sphere before anything else,
and three computes that sphere ONCE, from whatever instances existed then.
Nothing invalidates it when instances are added, moved or resized. A model
built node by node had its sphere computed while the mesh was empty, so every
later click was rejected before a single instance was tested — which is also
why every fixture-based test passed and the flow a person uses did not.

Two more faults in the same gesture: a node marker sized as a fraction of the
model diagonal with a 2 cm floor subtends two-thirds of ONE PIXEL on a 4 × 3 m
portal, and the marker is the pick target; and `nodeScreenPos` projected raw
coordinates for a model the application draws embedded flat in 3D.

## A raft is a structure

Modelling a mat foundation correctly means keeping the plates, giving them a
thickness and deleting the bars — and that produced a model PRO would not
solve. Five JS gates asked "are there frame elements" where they meant "does
anything carry load". None of it was the solver: the new test solves a
nine-node raft with ZERO elements and recovers a stress for every quad.

## Plates

Three separate things made the contour useless on the one model it exists for,
and none was the analysis. A face was two triangles, so four corner colours
across a shared diagonal gave a seam and a flat field between them — it is
subdivided 6 × 6 now, each vertex carrying the weight of each corner. The ramp
was white → red over a symmetric amplitude, so `mx` from −33 to 1735 kN·m/m
landed in the top decade of one hue and the plate came out uniform pale pink —
the scale is spent on the range the results occupy and walks blue → red, in
one function used by the painter AND the legend. And it opened on von Mises,
which for a slab in bending the application itself reports as negligible.

The deformed shape ignored shells entirely: a raft answered the deformation
slider with an empty screen.

## Drawing, and the ribbon

PRO could not place geometry with the pointer, and none of it needed building:
`Viewport3D` has placed nodes and built members from two clicks since 3D
existed, and PRO's ribbon set `currentTool` to exactly `select` and `pan`. One
Draw group now gives the table AND the tool, because typing a node's
coordinates and clicking a member onto it are one job, not two ways of working.

"Colour maps" is gone as a group — a colour map is a way of DRAWING a quantity,
which Basic has said for a while, and PRO now calls the same
`representationsFor` / `showQuantityAs`. The CIRSOC verification map went with
it: painting members by a code-check outcome is a design step. Stress became a
command that paints everything carrying load, with members and plates
switchable. Results left Inspect (every command above it already opens that
panel) and Diagnostics moved into Advanced, where the other things you ask the
model to do live.

## Curves

A curved SHELL was fully supported and had no control — the model has carried
`curved` since shells existed and the solver reads it. It is a checkbox now,
and the panel detects the case: whether four points are coplanar is arithmetic.

A curved MEMBER did not exist. The solver has straight elements and no curved
beam, so an arc is materialised as a chain of them — what every commercial
package does, and what lets diagrams, verification and detailing keep working.
Three points, because that is what an engineer has. The points are ON the arc,
and the panel states how far the chords still fall inside it. Existing nodes
are reused: an even segment count lands a generated point exactly on the node
that defined the curve, and two nodes in one place analyse as two nodes.

## The report, and the take-off

"Cómputo de materiales" reported no materials on a structure that had just
solved, because it asked whether the members had been CODE CHECKED. What a
structure is made of is a property of its geometry; `model-takeoff.ts` measures
members and shells, groups by material, and reports what it could not measure
instead of dropping it.

The report drops CIRSOC verification (a design output) and story drift (a
permanently disabled checkbox). Advanced analysis was one checkbox over as many
as six studies — it is itemised over the ones that actually ran, and a section
with everything deselected prints no heading rather than an empty one. Project
data is a disclosure and a blank letterhead prints nothing.

## Audited rather than assumed

All seventeen analyses in the Advanced panel were PRESSED on a solved model:
**11 answer, 6 refuse for a stated reason, 0 are silent.** Every refusal is a
missing input. `docs/proposals/pro-advanced-analyses.md` says what each does.

The first version of that sweep reported seven as silent and was wrong — it
anchored to the button's own block and missed answers printed one block over.
A measurement that mistakes where an answer is printed for the absence of one
indicts working code.

## Two things about the harness

`nodeScreenPos` was projecting through the 2D canvas transform in 3D — a
confident number with nothing to do with where the node is, which can fabricate
a failure and can fabricate a pass.

And a preview is no longer reused when the build it serves is older than the
source. The per-worktree port fixed collisions between checkouts; this fixes
the one a checkout does to itself, which cost a green Education suite against a
bundle that predated the effect breaking it.

## Checks

Typecheck holds, with the baseline one lower than it was. The solver is
untouched.

---

# Third round

## Stairs

A stair is an inclined waist slab, and the steps on it are dead load. Not a
stack of little slabs: every code check of a stair runs on the waist, and a
triangular fillet cast on top of it carries no bending across its own joints.
Modelling the steps as geometry would add stiffness the real stair does not
have and return a deflection too small, with nothing saying so.

The solver needed nothing — `inclined-shell.test.ts` already locked that a quad
with its top edge out of the XY plane solves and recovers stresses. What was
missing was a way to ask for one: a stair meant working out the top edge's z by
hand and typing four node ids.

Two ways in, because both are how people work: from the bottom edge you start
at, or by tilting the slab you already drew. The tilt keeps the plan outline
and raises only the far edge, which makes the tread a DERIVED quantity — so the
panel prints it, since a conversion that silently re-proportions a stair is one
the reader cannot check. The step load is `γ·(riser/2)·cos α`, spread over the
INCLINED area because that is the area a surface load is integrated over; the
waist is deliberately not in that number, because the solver already weighs it.

Comfort is warned about and never enforced. Blondel's `2r + t` is a rule of the
human stride, not a clause anyone can cite, and a stair inside an existing
shaft is sometimes the stair you have. A geometry that cannot EXIST is what
disables the button.

## A plate you can still change your mind about

The `curved` flag was settable while creating a quad and never after, so "is
this a cáscara" had to be answered before the geometry was on screen, and a
slab whose corner is later lifted out of plane had no way to say so. It is a
checkbox in the plates table now — in the row that was already printing `≈` for
one — and a panel control that acts on the selection and states how far out of
plane those quads actually are, which is the fact the decision rests on.

## The grid, and two highlights that meant two things

Both defects here were mine, from the camera-following rewrite.

`GridHelper(size, n)` places its lines at `centre + (i − n/2)·spacing`: with n
even they land on multiples of the spacing, with n odd they land half a cell
off. The division count is derived from the zoom, so its parity flipped as you
zoomed and the whole floor slid sideways. And the rebuild was triggered by the
camera having moved 8 % of its distance, which rebuilds on orbits that change
nothing and MISSES a zoom that crosses a spacing threshold by less — which is
why the grid looked like it refreshed only when a node was added. Adding one is
what forced the next rebuild. The layout is a value now, computed every frame
for a dozen arithmetic operations, and the grid is rebuilt when it changes.

Separately: the BOX on a top-bar button means "this is what the right-hand
panel is showing" and the coloured ICON means "this is what the pointer is
doing". For every command those coincide, which is why one highlight served.
For Move and Select they come apart, and one highlight is then claiming
something untrue. Move opens no panel and never takes a box.

## The load generators, audited

Checked cell by cell against the markdown of the printed editions.

CIRSOC 101 §2.3.2 combinations: correct, all seven with both exceptions and the
F and H rules. CIRSOC 102 wind: correct on everything checked — the exposure
constants, Kz reproducing 0,71 / 1,00 / 1,19 at 10 m, Kd, and (GCpi).

**Tabla 4.1 had one wrong cell.** A passenger-vehicle garage was 2,5 kN/m²;
the 2025 table prints 2, and art. 4.10.1 prints it again. 2,5 is the 2005
figure. Nothing pinned it, which is how a value survives an edition change.

**The permanent load came from nowhere at all** — five fixed rows inside a
dialog whose every other figure cites a clause. Tabla 3.1 is transcribed now:
sixty-two rows, some per m² and some per m³ (where a missing thickness is
refused rather than guessed), the (*) batten footnote as a flag rather than
something to remember, and §3.1.4 checked — the partition allowance is required
whether or not the partitions are drawn, and what is missing shows up as
nothing at all.

## INPRES-CIRSOC 103, the static method

The seismic coefficient was a number the reader typed, filled in by a 2005-era
model with Ca and Cv by SOIL rather than by spectral type, T3 fixed at 3 s
where the 2018 table gives 3, 5, 8 and 13, and R built from a ductility μ
through a ramp instead of read off Tabla 5.1. No clause anywhere in it.

Capítulo 6 is implemented from the regulation, as the lost-branch handoff
requires of any successor: the spectrum [3.1]–[3.4] with Tabla 3.1's Ca and Cv;
`Ta = Cr·H^x` with the `Cu·Ta` cap of [6.7], the clause that stops a soft model
from designing its own load away; C from [6.3]/[6.4] with BOTH floors, [6.5] in
zones 3 and 4 and [6.6] in 0, 1 and 2 which has no R in it; `Vo = C·W`; and the
height distribution of [6.11], with [6.12]/[6.13] moving a tenth of the shear
onto the top mass past `T > 2·T2` — a branch a typed coefficient cannot reach,
because nothing in it says what T is. All thirty-six rows of Tabla 5.1, and
Tabla 3.3's f1, which separates a warehouse from a flat.

Four things it refuses rather than approximates: zone 0, site class SF, the
Tabla 5.1 row that prints R as an expression on the wall layout, and any
building §2.7.3 hands to the dynamic methods. Finding those exposed a real
defect — the plan returned early only on ROLE gates, so a refusal discovered
while generating came back beside `outcome: 'READY'`, which the caller applies.

**The zoning map is deliberately not digitised.** Anexo A assigns zones
department by department across five zones and every province, in two columns.
A transcription nobody checked that puts a building in zone 2 instead of zone 4
halves its base shear with no symptom. The zone is an input; the annex is cited.

## Asking for a regulation from the row that names the load

One button for the whole project is right while a project is being set up and
wrong once it has its gravity loads and needs wind. A load case row asks for
ITS regulation now: the `§` on a row typed W opens the dialog with the wind
block on and in view. Lr and S get no button, because the generator has no
section for them and a button that opened one would be promising something that
is not there.

## What screenshotting found

I built the new panels and read none of them on screen. Three defects no test
would catch: the dialog's primary button still said "Seismic preview" for a
button that previews the whole plan; `.al-warn` carried no font size and so the
sentence explaining a disabled checkbox was the largest text in the dialog; and
the stair's riser box clipped its last digit, which on a riser is the
difference between two stairs.

## Deleted

The 2005 seismic model, once it had zero callers. A superseded implementation
of a seismic code sitting unused is not harmless: it is the first thing anyone
searching this tree for "seismic" finds, it is in `engine/` where you would
look, and the correct one is not.

## Checks

Typecheck holds at its baseline. The solver is untouched.


diegokingston:
## Second pass, over the five commits from today

### Tabla 3.1 — checked independently, all twelve agree

You wrote that you verified these by hand and got your own check wrong twice, which seemed worth a second pair of eyes. The published text is in the repo (`docs/codes/CIRSOC/markdown/inpres-cirsoc-103-i/capitulo-3-…`), so I read the table off the page and compared every cell:

| Type | Zone 4 | Zone 3 | Zone 2 | Zone 1 |
|---|---|---|---|---|
| 1 (SA/SB/SC) | 0,37 / 0,51 ✓ | 0,29 / 0,39 ✓ | 0,18 / 0,25 ✓ | 0,09 / 0,13 ✓ |
| 2 (SD) | 0,40 / 0,59 ✓ | 0,32 / 0,47 ✓ | 0,22 / 0,32 ✓ | 0,12 / 0,18 ✓ |
| 3 (SE) | 0,36 / 0,90 ✓ | 0,35 / 0,74 ✓ | 0,30 / 0,50 ✓ | 0,19 / 0,26 ✓ |

`nearFault = zone >= 3` is right too: the page prints the `Na` / `Nv` factors on zones 3 and 4 only, and leaves 1 and 2 as bare numbers.

### The `auto-loads` deletion checks out

"nothing calls it" is accurate. There is exactly one import of that module left on the branch — `scripts/build-cad-dxf-examples.ts`, taking `OCCUPANCY_TABLE` and `DEAD_LOAD_DEFAULTS`, which are precisely the two tables your message says remain. The other files that mention `computeSeismicStatic` and friends carry their own 2025 implementations under `codes/cirsoc103/`.

### One finding: the eleven dictionaries are wrong, not stale

On `autoLoad.preview` you wrote that the eleven unoffered dictionaries "still carry the old text and I have left them, because I cannot vouch for translations I cannot read."

The instinct is right and I think the conclusion inverts. They do not carry stale English — they carry *translated* versions of the old meaning: `지진 미리보기`, `地震预览`, both "seismic preview". So a Korean reader gets a button confidently naming an action it does not perform, in their own language, which reads as deliberate rather than as a gap.

And `tAt` falls back:

```ts
return (dict as any)[key] ?? (dicts.en as any)[key] ?? key;
```

So **deleting** the key from those eleven shows them your corrected English — not a raw key. That is worse than a good translation and better than a confident wrong one, and it requires vouching for nothing, which is exactly your stated reason for not touching them.

And that path is already load-bearing right here: the two sibling keys you added, `autoLoad.previewTitle` and `autoLoad.previewStale`, exist **only** in en/es/pt. All eleven others already fall through to English for them today. So deleting `autoLoad.preview` from the eleven does not introduce a new state — it puts one key into the state its own two neighbours are already in.

Not urgent, and entirely your call; I mention it because "left untranslated" and "left saying something false" are different states and this is the second.

### Also read

- `clear()` now invalidates the node bounds, pushed earlier as `8e6d1762`, with the bounds test extended to walk every mutator rather than name one per case.
- The retraction in `04808952` marks the proposal rather than rewriting it — "a proposal edited to match what was built stops being evidence of anything" is the right call, and the reason the rest of that document is still worth reading.
- `85faeeb1` is the commit I would point at if someone asked what this branch does well: three defects that no test would have caught, found by looking at the screen, and said plainly.

Still not reviewed: the 16 `pro/` components in full, and the e2e specs beyond the camera ones.

diegokingston:
## Third pass: the `pro/` components and the e2e

Read the 20 changed files under `src/components/pro/` + `src/lib/pro/` (+2936/−777) and the 39 e2e specs (+2509/−108). Eight things worth acting on, then what I checked and cleared.

### 1. `overflow` is now dead machinery, and it fails silently

`ProRibbon` deleted the Select dropdown, which was the only code that ever set `openMenu` to a group id. `openMenu` is now assigned in exactly three places and all three are `null` (declaration, line 164, line 189). So this filter can never be true on its right-hand side:

```svelte
{#each g.cmds.filter(c => !c.overflow || openMenu === g.id) as c (c.id)}
```

Nothing is hidden today because no command sets `overflow: true`. But the field is still in the exported `ProCmd` type with a doc comment promising behaviour — *"Shown only when the group is expanded. Desktop ribbon only."* — that no longer has an implementation. The next person who follows that comment gets a button that never renders, with no error to tell them why.

Five CSS selectors went dead with it: `.pr-caret`, `.pr-menu`, `.pr-menu-item`, and — worth a second look — `.pr-menu-check` and `.pr-menu-sep`, which **this PR added**, for a menu that does not exist in this file.

Either drop the field, the filter and the CSS together, or give a group a way to expand again.

### 2. `repeatSelection` has no ceiling on `count`

`repeatIsMeaningful` guards the bottom carefully and the top not at all:

```ts
if (!Number.isFinite(spec.count) || spec.count < 1) return false;
```

`max="200"` on the input is markup-only — with `bind:value`, typing `50000` sets `count = 50000`. Then `for (let k = 1; k <= spec.count; k++)` runs that many times over the whole selection, inside one `modelStore.batch`, after `historyStore.pushState()` has already snapshotted. With a storey of 40 nodes selected that is two million `addNode` calls and a locked tab.

What makes this easy to call a defect rather than a judgement call is that **the fix already exists in this branch**, three files away, with the reasoning written out:

```ts
/* … Bounded so a mistyped step count cannot mesh the model to a halt. */
const divisions = $derived(Math.min(40, Math.max(1, Math.round(steps))));
```

Same failure, same author, same PR — clamped in the stair tool, unclamped in the repeat tool. One `Math.min` in `repeatIsMeaningful` (where the lower bound already lives) closes it.

### 3. The shell recommendation never updates on the pick path

`updateNewRecommendation()` is called from exactly two places: the corner boxes' `oninput`, and the thickness input's. The `$effect` that mirrors viewport picks into the boxes does not call it:

```ts
$effect(() => {
  const p = uiStore.shellNodePick;
  if (!p.active) return;
  const next = ['', '', '', ''];
  p.picked.forEach((id, i) => { if (i < 4) next[i] = String(id); });
  newNodeIds = next;                      // ← no recommendation refresh
});
```

The old code did refresh there — the deleted branch read `plateNodes = [...]; updatePlateRecommendation();`. So the recommendation box is blank on precisely the flow this PR is built to promote: press "Draw plate in the model", click three corners, get nothing. Type the same ids by hand and it appears.

`newPts` is already `$derived`, so `newRecommendation` can be `$derived.by` and the whole class of bug goes away.

### 4. The formulation override is gone, and the comment says it isn't

The creator's comment closes with:

> The formulation is still shown, and still overridable for someone who wants to, under the detail.

Shown, yes. Overridable, no: there is no `newFamily` state, no select, and `shellFamily` is never assigned anywhere in the new file — the old `addPlate`/`addQuad` both set it from the `auto`-or-override choice.

Worth knowing before deciding what to do: **the solver never reads `shellFamily`.** The only consumers in the repo are `url-sharing.ts` (round-trip) and its tests. So nothing computes differently, and DKMT / MITC9 / SHB8PS recommendations were already advisory. The loss is the control and the accuracy of the comment, not the analysis — but the comment should say whichever is true.

### 5. `aria-pressed` means two different things on two adjacent buttons

The two-channel visual design is well argued and I am not questioning it — the box for "this is the panel showing", the coloured icon for "this is what the pointer is doing". The problem is that ARIA got one channel and it was assigned to the one that does not describe the pointer:

```svelte
<button data-testid="pr-pan"    aria-pressed={uiStore.currentTool === 'pan'} …>
<button data-testid="pr-select" aria-pressed={selectPanelShowing} …>
```

Two buttons, side by side, same attribute, different properties. The case the design is proudest of is the one that reads worst: in select mode with the Nodes panel open, a screen reader says *"Select, not pressed"* while select is exactly what the pointer is doing. `pro-tools-and-panels.spec.ts:44` pins that state, so this is a deliberate choice rather than an oversight — which is why I am raising it as a disagreement and not a bug.

Two things point the other way. `DrawInModelButton`, added in this same PR, uses `aria-pressed` for the tool. And `aria-current` exists for exactly "this is the one currently showing". Mapping `aria-pressed` → tool on both buttons and `aria-current` → panel gives the non-visual layer the same two channels the visual one has.

### 6. `pro-advanced-audit` can record a broken analysis as "answered"

The detector is "the whole `.adv-tab` text changed", polled every 500 ms:

```ts
const before = (await tab.innerText()).trim();
await btn.click();
for (let waited = 0; waited < 20_000; waited += 500) {
  await page.waitForTimeout(500);
  if ((await tab.innerText()).trim() !== before) { answered.push(name); return; }
```

Any text that appears before the analysis finishes satisfies it — a spinner, a progress line, an elapsed-time counter. An analysis that prints "Calculando…" and then throws inside the worker is recorded as having answered. That is the exact failure this suite was written to catch, and it is the one input that can slip past it. Comparing against a terminal state, or excluding the known in-progress markers, would close it.

### 7. …and `blocked` is collected but never asserted

Same test. `blocked` is filled, logged, and then only `silent` and a count are checked:

```ts
expect(silent, `…`).toEqual([]);
expect(answered.length, 'several actually ran').toBeGreaterThan(3);
```

Against sixteen analyses, "more than three" means **twelve could quietly become disabled and this still passes** — which is precisely what a regression in an `enabled: () => solved` predicate would look like, and this PR moves several of those around. The data is already in hand; asserting `blocked` against an expected set by name turns a log line into a guard, and makes a newly-disabled analysis a failure rather than a thing someone notices in the console six weeks later.

### 8. Repeat drops the loads, and does not say so

`RepeatSource` carries nodes, elements and supports. The word "load" does not appear anywhere in `array-copy.ts`.

For a tool whose headline case is "six identical storeys", the loads on those beams are the most tedious thing left to redo by hand. What makes it read as an oversight rather than a decision is that the module's *"What this does not do"* section names the other omission explicitly and at length — it will not weld coincident nodes, and says why. Silence about loads next to a paragraph about welding invites the reader to assume loads come along. There is a `withSupports` checkbox and no `withLoads`.

If it is deliberate, the doc-comment section it belongs in already exists.

### Minor

- **`ProDeadLoadBuilder` keys rows by index** — `{#each rows as row, i (i)}` with a `remove(i)` that splices. Values re-read correctly from `rows[i]`, so nothing displays wrong; what transfers to the neighbouring row is DOM state — focus, and a half-typed number the browser has not committed to the binding yet.
- **A gap in the shell corner boxes misreports itself.** `newCount` counts non-empty boxes but `slice(0, newCount)` takes the *first* N. Fill N1, N2 and N4 and it reads N3 — empty — so `addShell` fails with `pro.errNodesExist` ("nodes must exist") while the value in N4 is quietly ignored. Fails safe, explains itself wrongly.
- **`${t('pro.drawInModel')} ${label}`** builds a sentence from two separately-translated fragments, in an app shipping 14 locales. No translator can reorder it. `tp('pro.drawInModelX', { what: label })` would let them.
- **`openFirstExercise` uses `count()` to decide whether to click.** `count()` does not retry, so it reads panel state at one arbitrary instant after `leaveExercise()`. Not a defect I can demonstrate, but it is where this kind of suite usually starts flaking.

### Checked and cleared

Recording these so nobody re-does them — each looked like a finding and wasn't:

- **The `maps` and `inspect` groups being deleted from Analyse.** The verification colouring is still reachable (`ProVerificationTab.svelte:662` sets `diagramType = 'verification'`, where the verification results actually live — arguably better than the ribbon). `results` is the Analyse stage's `home`, and `diagnostics` is reachable from `ProDesignTab`, `ProPanel` and a section inside Results.
- **The Select dropdown being deleted.** `SelectionPanel` does carry the kind picker, via `selectsKind` / `toggleSelectKind` — and it is strictly richer than what it replaced, because it does multi-kind. The ribbon comment is accurate.
- **`void downloadExcel()` discarding a promise.** Safe by design: `loadXlsxModule` catches the dynamic-import failure, toasts, and returns null, with a comment that anticipates this exact review. Nothing rejects into the void.
- **`document.querySelector('canvas')` in `report-export.ts`.** `Viewport` and `Viewport3D` are an `{#if}/{:else}` in `App.svelte`, so exactly one viewport canvas is ever mounted.
- **The link members in `array-copy`** guard the no-elements-selected case (`if (proto)`), and the "chain, not a fan" behaviour is pinned by its own test.
- **e2e tagging.** All fourteen new specs carry `@smoke` except `pro-advanced-audit`, which is `@slow` — so it runs on main or behind the `run-e2e` label, not on this PR by default. That matches the convention; flagging only so the audit's result is not assumed to be part of this PR's green tick.

The e2e work in this PR is the strongest part of it. `edu-in-basic`'s note on clearing storage once rather than per-navigation, and `pro-advanced-audit`'s on anchoring to the whole panel because seven analyses print into a sibling, are both harness bugs that would have produced confident wrong results — found and written down rather than worked around.

diegokingston:
## Fixes for the review above

Pushed. Seven of the items, the three that are yours to decide left alone and said so below.

### Fixed

**`overflow`, and the menu machinery under it.** Removed `openMenu`, the filter it gated, the `overflow` field, and the six dead selectors (`.pr-dd`, `.pr-menu`, `.pr-menu-item`, `.pr-menu-check`, `.pr-menu-sep`, `.pr-caret`). A note on `ProCmd` says what the flag meant and what would have to come back with it, so the next person who wants an overflowing group builds the affordance before the flag.

**A ceiling on `repeatSelection`.** `MAX_REPEAT_COPIES = 200`, refused in `repeatIsMeaningful` where the lower bound already lived, and the panel's `max` now reads from the same constant so the bound a reader sees and the bound that holds are one bound. Two tests: the ceiling itself, and that nothing is built when a count over it is asked for anyway. The panel says why the button is grey rather than only greying it — `repeat.tooMany`, en/es/pt, falling through to English elsewhere.

**The shell recommendation, on the pick path.** `newRecommendation` is `$derived.by` now instead of a `$state` refreshed from two `oninput` handlers, so it follows the corners however they arrive. The manual refresh is gone from both inputs.

**The gap in the corner boxes.** `newGiven` takes the corners that were actually given rather than the first N boxes, so filling N1, N2 and N4 uses N4 instead of silently reading the empty N3 and failing with "those nodes do not exist".

**The formulation comment.** Rewritten to say what is true: shown, not chooseable, and the `auto` value the old creators wrote is not written either. It also records the thing that decides how much any of that matters — `shellFamily` is read by nothing in the engine, only by `url-sharing.ts` — so whoever picks this up starts from the fact rather than rediscovering it.

**Loads, in "What this does not do".** `array-copy.ts` named one omission at length and left the other silent, which reads as a complete list. Now it names both. Whether Repeat should carry loads is still yours; the doc no longer implies it does.

**`ProDeadLoadBuilder` rows keyed by the row.** `remove` filters and both adders append, so identity survives all three — no interface change needed, and focus stops moving to the neighbouring row on delete.

### One fix I could not verify, and what could bite

`pro-advanced-audit` now waits for the panel to come to **rest** rather than to **change**: text different from before, unchanged across two further polls, and the button no longer held disabled by `solving`. The hole was concrete — about twelve of these buttons carry `{solving ? t('pro.solving') : …}`, and that label is inside the `.adv-tab` this reads, so the first poll after the click always saw different text.

**It is `@slow`, so CI will not run it on this PR without the `run-e2e` label, and I have not run it either.** The failure mode to watch for, if you do label it: an analysis whose button does *not* return to enabled after its run would now be recorded as silent rather than answered. I checked the preconditions on all sixteen and none are cleared by running the analysis, and `handleSectionAnalysis` — the one button not gated on `solving` — has no busy label to confuse the settle check. But that is reading, not a run.

### Left alone, deliberately

**`aria-pressed` on Select.** Pinned by `pro-tools-and-panels.spec.ts:44`, so it is a decision and not an oversight, and changing it means changing that test. My argument is in the review above; the short version is that `aria-current` exists for "this is the one showing" and `DrawInModelButton` already uses `aria-pressed` for the tool. Your call.

**`blocked` in the advanced audit.** The right assertion is the expected blocked set by name, and I am not going to guess it — a wrong list breaks CI on main, which is worse than the gap. It needs one labelled run to read off, and then it is three lines. Happy to do that if you want to kick off `run-e2e`.

**`${t('pro.drawInModel')} ${label}`.** A design suggestion, not a defect: it builds a sentence from two separately-translated fragments and no translator can reorder it. Fixing it properly means a placeholder key, which touches the dictionaries, and after the `autoLoad.preview` conversation I would rather you decide how much churn there is appetite for.

### Gate

`npm run typecheck` — 472, baseline 472, no new errors. `npm run test` green. The three type errors I saw first were an unbuilt WASM in a fresh worktree, not the changes.

### One correction to myself

Mid-way through this I thought the whole `repeat.*` namespace was undefined and that the Repeat panel was rendering raw keys. It is not — the keys are all there in en/es/pt. My grep had run from the wrong directory and returned empty because the path did not exist, not because there were no matches. Nothing in the review above depended on it, and I checked the rest of the branch properly afterwards: of 3,668 literal keys asked for across the components, every one resolves, with the only six "missing" being prefixes built by concatenation.

